Issue #8, opened 2022-09-24
granular measurements


DarwinsBuddy:
Instead of having a total sensor per Zählpunkt, it would be nice to have a sensor capable of providing the same granularity in measurements as WN delivers. Unfortunately, WN delivers them not as they appear, but merely once a day. 

Find the right way in HomeAssistant to store those measurements in bulk for the individual 15minutes/hourly intervals per day, when the API is returning them finally.

DarwinsBuddy:
see https://github.com/DarwinsBuddy/WienerNetzeSmartmeter/issues/4#issuecomment-1256890357

gluckb:
I am just thinking...
Either way the consumtion is not assigned to the right day, because we get them one day later.
As you wrote, with 15Min opt-in we get a list of quarterly values at around 2:00 am. Would it be possible to read the values corresponding to the actual system time and feed the sensor with the sum of values between 0:00 and actual time?

e.g.:
response to (..)verbrauch?period=DAY&accumulate=False&offset=0&dayViewResolution=QUARTER-HOUR&dateFrom=2022-09-30T00%3A00%3A00.000Z'

"yesterdays_consumption"{
   "quarter-hour-opt-in":true,
   "values":[
      {
         "value":55,
         "timestamp":"2022-09-30T00:00:00.000Z",
         "isEstimated":false
      },
      {
         "value":62,
         "timestamp":"2022-09-30T00:15:00.000Z",
         "isEstimated":false
      },
(..)

data processing idea:
(data unavailable until 2:00 am)
at 2:00 am the sensor gets the sum between 0:00-2:00, also 55 + 62 + (..)
at 2:15 am the sensor gets the sum between 0:00-2:15
at 2:30 am the sensor gets the sum between 0:00-2:30
and so on.

I think this would also fit to the sensor class TOTAL_INCREASING.

I also noticed, that the values you export as csv via Smart Meter WebPortal are shifted in time, when you compare to the API response or the Smart Meter WebPortal Graph. In my case:

date | value | csv timestamp | API timestamp | WebPortal Graph timestamp
30.09.2022 | 0,062 | 00:00:00.000 | x | x
30.09.2022 | 0,058 | 00:15:00.000 | x | 00:00:00.000
30.09.2022 | 0,068 | 00:30:00.000 | x | 00:15:00.000
30.09.2022 | 0,058 | 00:45:00.000 | x | 00:30:00.000
[...]

PR #20, opened 2022-11-26
[WIP] Proof of Concept for importing hourly consumption

reox:
This imports the hourly consumption into the statistics entity. If the integration is set-up for the first time, the current meter reading is fetched and used as a baseline.

As discussed in #8, this is a proof of concept... I'm not experienced with the HA API, so maybe some things are stupid 😅
Also there are a lot of TODO in the code, where I did not have a good answer for...


reox:
Found the first problem already: After some minutes of the first import, the statistics will get to a negative number and say that energy went into the grid.... Possibly setting the initial meter reading is not the best idea...

reox:
And there is a second problem: For some reason, it imported the same date twice. Is it possible that the API returns the whole day irrespective of the time?

custom_components/wnsm/sensor.py
```diff
@@ -157,6 +175,76 @@ def parse_quarterly_consumption_response(self, response):
...
+        while iterations > 0 and not has_none and start < now:
+            _LOGGER.debug(f"Select 24h of Data, using sum={sum_:.3f}, start={start}")
+            iterations -= 1
+            verbrauch = await self.get_consumption(smartmeter, start)
+            _LOGGER.debug(verbrauch)
+
+            # Check if this batch of data is valid and contains hourly statistics:
+            if not verbrauch.get('quarter-hour-opt-in'):
+                _LOGGER.warning(f"Data starting at {start} does not contain granular data! Opt-in was not set back then.")
+                start += timedelta(hours=24)  # Select the next day...
```
DarwinsBuddy, line 212: is this necessary? shouldn't the `end_date` of the request already be set to `datetime.now()` by default, thus returning all consumption data beginning from `start` till now? (we could also paginate through the days by incorporating above mentioned additional `end_date` param, but if it's possible to get all data at one request, I guess this saves a lot of time)

custom_components/wnsm/sensor.py
```diff
@@ -157,6 +175,76 @@ def parse_quarterly_consumption_response(self, response):
...
+                continue
+
+            if 'values' not in verbrauch:
+                _LOGGER.error("No values in API response!")
```
DarwinsBuddy, line 216: if it's an error, shouldn't we abort? or is it okay to lose some "days" worth of data?

custom_components/wnsm/sensor.py
```diff
@@ -157,6 +175,76 @@ def parse_quarterly_consumption_response(self, response):
...
+                continue
+
+            # TODO: What happens on summer-/wintertime change in the statistics?
```
DarwinsBuddy, line 219: when converted to utc, I guess we do not have to worry about that. or am I mistaken?

custom_components/wnsm/sensor.py
```diff
@@ -157,6 +175,76 @@ def parse_quarterly_consumption_response(self, response):
...
+            for v in verbrauch['values']:
+                # Timestamp has to be aware of timezone
+                ts = datetime.fromisoformat(v['timestamp'][:-1]).replace(tzinfo=timezone.utc)
+                if ts < start:
+                    _LOGGER.debug(f"Timestamp from API ({ts}) is less than start time ({start}), ignoring value!")
+                    continue
+                if ts.minute != 0:
+                    _LOGGER.error("Minute of timestamp is non-zero, this must not happen!")
+                    return
+                if v['value'] is None:
```
DarwinsBuddy, line 229: as of my knowledge, unfortunately yes. I've seen a lot of data points missing in between (mostly in the morning hours 0-3). These have to be interpolated I guess.

custom_components/wnsm/sensor.py
```diff
@@ -157,6 +175,76 @@ def parse_quarterly_consumption_response(self, response):
...
+                    # If not, we could break as soon as we hit the first None value.
+                    has_none = True
+                    continue
+                elif has_none:
+                    _LOGGER.warning("Value is suddenly not None anymore!")
+                sum_ += Decimal(v['value'] / 1000.0)  # Convert to kWh, and accumulate
+                statistics.append(StatisticData(start=ts, sum=sum_))
+
+                # Set new start date for next batch
+                start = ts + timedelta(hours=1)
```
DarwinsBuddy, line 241: haven't we already incremented start by 1 day prior to this?

custom_components/wnsm/sensor.py
```diff
@@ -166,29 +254,72 @@ async def async_update(self):
...
-            # TODO: find out how to use quarterly data in another sensor
-            #       wiener smartmeter does not expose them quarterly, but daily :/
-            #self.attrs = self.parse_quarterly_consumption_response(response)
-            if self.is_active(zp):
+
+            if not self.is_active(zp):
+                _LOGGER.warning("Smartmeter not active...")
+                return
+
+            # TODO: why does self.entity_id returns None?
```
DarwinsBuddy, line 262: does it? it should return the entity's id like in our case `sensor.<zaehlpunktnummer>` :thinking:

custom_components/wnsm/sensor.py
```diff
@@ -166,29 +254,72 @@ async def async_update(self):
...
+            # From
+            # https://github.com/DarkC35/ha_linznetz/blob/904f361e760103f900ad93522a0215d348fc83bb/custom_components/linznetz/sensor.py
+            # Select one entry from the statistics, convert the units
+            last_inserted_stat = await get_instance(self.hass).async_add_executor_job(get_last_statistics, self.hass, 1, entity_id, True)
+            _LOGGER.debug(f"Last inserted stat: {last_inserted_stat}")
+
+            if len(last_inserted_stat) == 0 or len(last_inserted_stat[entity_id]) == 0:
+                # No previous data
+
+                # FIXME: This seems not to work and after some time you get a negative consumption
```
DarwinsBuddy, line 275: could that be related to the example of the "linz" sensor being of type `SensorStateClass.TOTAL`
and this here being of type `SensorStateClass.TOTAL_INCREASING` ?

total increasing may only be incremented. maybe belatedly updating it causes internal discrepancies. we could think of changing the sensor to be a total one. but for downwards-compatibility it may be wise to add a sensor capable of this granularity of type `TOTAL` (if that's the culprit)

DarwinsBuddy (commented):
I tried my best to share some thought about it, although it was not easy to follow all of your steps. But I am pretty sure you are on the right track there. Thanks a lot for taking this effort. Let me know when I can be of any help.

custom_components/wnsm/api/client.py
```diff
@@ -109,15 +109,15 @@ def _call_api(
         if query:
             url += ("?" if "?" not in endpoint else "&") + parse.urlencode(query)
 
-        logger.debug("REQUEST: {}", url)
```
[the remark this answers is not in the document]
reox: Is there any advantage of using the formatting in the logger? I use both variants all the time... But yes, format strings look better!

custom_components/wnsm/api/client.py
```diff
@@ -179,10 +179,12 @@ def verbrauch(self, date_from: datetime, date_to: datetime = None, zaehlpunkt =
                 'm/messdaten/zaehlpunkt/ZAEHLPUNKT/verbrauch'
         """
         if date_to is None:
-            date_to = datetime.now()
+            date_to = date_from + timedelta(hours=24)  # You can not get more than that anyways it seems...
         if zaehlpunkt is None:
             zaehlpunkt = self._get_first_zaehlpunkt()
         endpoint = "messdaten/zaehlpunkt/{}/verbrauch".format(zaehlpunkt)
+        # FIXME: If period=DAY is set in the args, you will get a maximum of 24h, irrespective of
```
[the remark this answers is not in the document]
reox: Yes, I think too that it is a limitation or I did something wrong... I haven't checked if you can query less than 24h though - maybe the end date would be useful for that?

custom_components/wnsm/sensor.py
```diff
@@ -157,6 +175,76 @@ def parse_quarterly_consumption_response(self, response):
         self._state = sum
         return data
 
+    async def _import_statistics(self, smartmeter: Smartmeter, start: datetime, sum_: Decimal):
+        """Import hourly consumption data into the statistics module, using start date and sum"""
+        if start.tzinfo is None:
```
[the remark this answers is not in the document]
reox: I think there are many different ways how to setup hassio. I for example use HAOS and I think there the time is set to UTC on the host system - because you only set the timezone in the frontend...
The API is UTC and also the statistics table is UTC - the only difference is, that the statistics module requires the tzinfo to be set and the API ignores it (right?)
I think the easiest is to always use UTC timezone, for example by using `datetime.datetime.now(datetime.timezone.utc)`, right?

custom_components/wnsm/api/client.py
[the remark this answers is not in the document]
DarwinsBuddy: No advantage except that it's more comprehensible imho :)

custom_components/wnsm/sensor.py
reox: First of all, the whole loop needs to be rewritten... I had some endless looping going on :sweat_smile: 
maybe a better way would be to iterate over tuples of `(start_date, end_date)` and thus knowing what the next `start_date` is beforehand? 
I also noticed, that if for some reason the loop fails and no data is inserted, the same `start_date` will be used over and over again, until something is written to the statistics table.
Maybe there is a better way of doing that, for example by inserting dummy data for date ranges where no consumption was recorded (for example because there was no opt-in?)
Another question would be, if you can select daily consumptions? Do you have to specify something like "WEEK"? I checked on the Smartmeter Webportal and I can not see a request to the `verbrauch` endpoint when selecting the Week, Month, or Year view? Only for Year it uses the `verbrauchRaw` with the parameter `granularity=DAY`. But where do the week and month stats come from?

custom_components/wnsm/api/client.py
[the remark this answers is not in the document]
DarwinsBuddy: I think 24h fits quite nice, but it's merely a parameter more if you e.g. want to query 3 days at once. (which will be the case anyways iirc when you pass a start date today-3days)

custom_components/wnsm/sensor.py
reox: True, Idk though how the API behaves. On an error, will it just not send any values (thus in this if you should raise an exception)?

custom_components/wnsm/sensor.py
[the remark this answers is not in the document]
DarwinsBuddy: aaah I see. well then we should use utc everywhere not to interfere with any hassio internals. I'm using the supervised version and was not sure though. Let's stick to UTC then.

custom_components/wnsm/sensor.py
reox: For the "turn the clock one hour back day" you will have the same time twice. That could be an issue, if you add the same time twice, thus I don't know how the API response is in that case...
Actually, one would need to query the 30th October to check - but my smartmeter wasn't in opt-in mode yet for that time...

custom_components/wnsm/sensor.py
reox: But for interpolated data, does it not write `isEstimated: True`? I checked my data and could not find one, but maybe I have to download the csv and check there... I guess `None` simply means "No value recorded so far"?

custom_components/wnsm/sensor.py
reox: No, that is only incremented by 24h if the batch does not contain opt-in data. L213 has a continue

custom_components/wnsm/sensor.py
reox: For me it returns `None` :thinking:
Maybe I have to check again...

custom_components/wnsm/sensor.py
reox: That could be the case. But I'm also not entirely sure how it happened. I noticed it twice in the docker image I used for testing, that after some time I got a negative peak. But only in the statistics - which is probably unrelated to the state class? But I have to admit, I do not know all the details of hassio sensors by heart...

custom_components/wnsm/sensor.py
reox: I poked the API a bit and it looks like on error you get an http error 500 and only this basic json:
```
{'quarter-hour-opt-in': False,
 'statistics': {'average': 0, 'maximum': 0, 'minimum': 0}}
```
Thus, checking if there are `values` seems a method to check whenever the API had an error, thus it is probably good to exit the function here

custom_components/wnsm/api/client.py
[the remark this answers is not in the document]
reox: It looks like you can change the dateTo to whatever you want and will always get 24h of data... Can you replicate that? Thus, this function can be much simplified by removing the dateTo...

custom_components/wnsm/sensor.py
reox: Nope, can not get it to work... `self.entity_id` is None for me...
Is there something you need to set in the constructor of the class, so that `entity_id` is populated?

custom_components/wnsm/sensor.py
```diff
@@ -130,6 +141,13 @@ async def get_daily_consumption(self, smartmeter: Smartmeter, date: datetime):
         else:
             return response
 
+    async def get_consumption(self, smartmeter: Smartmeter, start_date: datetime):
+        """Return the consumption starting from a date"""
+        response = await self.hass.async_add_executor_job(smartmeter.verbrauch, start_date, None, self.zaehlpunkt)
```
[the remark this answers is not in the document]
reox: as far as I understand that, an end date only works for the `verbrauchRaw` endpoint, to query the daily consumption.
I removed the end dates from the functions that use `verbrauch`

custom_components/wnsm/sensor.py
reox: I added another check for estimated data now and only skipped None readings. So far, I did not had this in my readings, so I'm not sure what to expect there...
If a None value pops up between actual readings now, it would be ignored and never be fetched again, because as far as I understand the statistics module, it is not possible to insert values in between.

custom_components/wnsm/sensor.py
DarwinsBuddy: agree, but better describe this behaviour here in a comment and... I guess it's time to start a document where we are summarizing these quirks of the API: https://github.com/DarwinsBuddy/WienerNetzeSmartmeter/discussions/25

custom_components/wnsm/sensor.py
```diff
@@ -157,6 +175,76 @@ def parse_quarterly_consumption_response(self, response):
         self._state = sum
         return data
 
+    async def _import_statistics(self, smartmeter: Smartmeter, start: datetime, sum_: Decimal):
+        """Import hourly consumption data into the statistics module, using start date and sum"""
+        if start.tzinfo is None:
+            raise ValueError("start datetime must be timezone-aware!")
+        # Have to be sure that full minutes are used. otherwise, the API returns a different
```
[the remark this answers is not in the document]
DarwinsBuddy: @reox Better pull the input sanitzing into the `get_consumption` function so we have that culprit measurement in one place, and document it there, s.t. we can reference it in https://github.com/DarwinsBuddy/WienerNetzeSmartmeter/discussions/25

custom_components/wnsm/api/client.py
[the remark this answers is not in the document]
DarwinsBuddy: unfortunately not, I don't get anything back right now. I'm a bit puzzled, since my hassio integration works as expected.

```In [29]: s.verbrauch(before(before(before(now))), zaehlpunkt=id)
Out[29]: 
{'quarter-hour-opt-in': False,
 'statistics': {'maximum': 0, 'minimum': 0, 'average': 0}}

In [30]: s.verbrauch(before(before(now)), zaehlpunkt=id)
Out[30]: 
{'quarter-hour-opt-in': False,
 'statistics': {'maximum': 0, 'minimum': 0, 'average': 0}}

In [31]: s.verbrauch(before(now), zaehlpunkt=id)
Out[31]: 
{'quarter-hour-opt-in': False,
 'statistics': {'maximum': 0, 'minimum': 0, 'average': 0}}

In [32]: s.verbrauch(now, zaehlpunkt=id)
Out[32]: 
{'quarter-hour-opt-in': False,
 'statistics': {'maximum': 0, 'minimum': 0, 'average': 0}}
```

custom_components/wnsm/sensor.py
DarwinsBuddy: 1. the iterating over tuples makes sense
2. when implementing pagination we have to keep in mind, that we are on a rogue mission here. the api could be hammered by all of our users, thus if we are not carefully using it by e.g. restricting it to merely 2-3 calls per update, it could be that WienerNetze will change their API so we cannot use it any longer, due to their concern of having outages. Let's bear that in mind when working on that.
3. the dummy-data I'd replace by the average consumption for the timerange queried. usually we are getting his back and if not we can for sure calculate the average ourselves and insert it. an even better solution would be to store the values as is (also with the gaps) and fill them in a later time, when they are queried anew - e.g. the next day or the one after that, at most - when another update happens and maybe the gaps where filled in by Wiener Netze belatedly.
4. I had a look at the API and I have following hypothesis how this API<->frontend intergation works

"Tag 1/4 h" triggers `/ereignisse` with `dayViewResolution=QUARTER-HOUR`
"Tag 1 h" triggers `/ereignisse` with `dayViewResolution=HOUR`
"Woche" here it's speculation, cause there is no API call I can spot that returns the initial data. So I'd guesstimate that this is served pre-rendered into the html by the server (which I have to proof somehow by finding it in the html upon load). But I could also overlook something.
"Monat" the same data is taken into account for that action, as no other calls than `/ereignisse` is taking place. so It's just aggregated differently by the javascript
"Jahr" that's going for `/verbrauchRaw` which is returning the daily consumptions. actually before digging into the odd behaviour of "Woche"/"Monat" I'd suggest to just take "/verbrauchRaw" and aggregate ourselves.

`/ereignisse` get queried on every action

custom_components/wnsm/sensor.py
DarwinsBuddy: I am also not quite sure how often this happens. What's your opinion on that:

on an error that is user specific (like wrong configuration, changed credentials, etc.) we should raise an exception.
on something that could be a flake or a hiccup on API or network side, we just log it as an error and proceed.

so in this case: maybe log an error and abort the update?

custom_components/wnsm/sensor.py
DarwinsBuddy: ah I see. that also answers my comment from above (regarding filling in the gaps in the upcoming days). well I guess that's fine as you did it, since we cannot compensate the crappy API in every matter.

custom_components/wnsm/sensor.py
DarwinsBuddy: what about `self._id` ?

seen at: https://community.home-assistant.io/t/custom-component-how-to-get-the-current-value-for-a-sensor/273608

custom_components/wnsm/sensor.py
DarwinsBuddy: sure. I'd suggest let's focus on the above discussions and try to get it to work regarding the API, afterwards we can (if it still persists) focus on this quirk :)

DarwinsBuddy:
made some comments on the PR, feel free to review it whenever you have time. :)

custom_components/wnsm/sensor.py
reox: > when implementing pagination we have to keep in mind, that we are on a rogue mission here. the api could be hammered by all of our users, thus if we are not carefully using it by e.g. restricting it to merely 2-3 calls per update, it could be that WienerNetze will change their API so we cannot use it any longer, due to their concern of having outages. Let's bear that in mind when working on that.

yep, so the easiest would be to remove the loop completely and always select a single day. 
The loading of historic data, i.e. the last month, is nice for testing, but if we just grab the last 48h for the beginning, it would reduce the traffic significantly.
Furthermore, one could increase the `SCAN_INTERVAL` to 1 or 2h.

DarwinsBuddy: I like that. if there is the need we can still extend it or provide the possibility to load historical data on demand

reox: An interesting approach would be to make it possible to upload the CSV file with the quarter hour statistics when setting up the integration.
There seems to be a specific API endpoint to export everything at once too:
```
https://api.wstw.at/gateway/WN_SMART_METER_PORTAL_API_B2B/1.0/zaehlpunkte/messwerte?zaehlpunkt=XXX&datumVon=2019-12-04&datumBis=2022-12-04&wertetyp=QUARTER_HOUR
```
This is used in the CSV export.
Funny enough, why is this the business2business API and not business2consumer? 🤔

reox: I hesitate to remove the loop though: It is tricky to decide if there are no values stored (and never will be) or if there are no values yet. Thus, if we not got through all values up to now, we might get stuck on a day where no values are stored (and never will be). In such a case, it would be guessing if we should skip the day or if we need to wait until the database is populated.

custom_components/wnsm/sensor.py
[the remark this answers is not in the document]
reox: Actually, it is not so much a quirk with the API this time but with HA: it requires that the datetime is timezone aware, to insert it in the statistics. And because I need to compare the datetime with other datetimes, I need to have all of them timezone aware...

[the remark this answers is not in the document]
DarwinsBuddy: I see. then I got that wrong. thanks for clarifying

custom_components/wnsm/sensor.py
DarwinsBuddy: nice find. I'll create an issue to pull this one out for later. 

> An interesting approach would be to make it possible to upload the CSV file with the quarter hour statistics when setting up the integration. There seems to be a specific API endpoint to export everything at once too:
> 
> ```
> https://api.wstw.at/gateway/WN_SMART_METER_PORTAL_API_B2B/1.0/zaehlpunkte/messwerte?zaehlpunkt=XXX&datumVon=2019-12-04&datumBis=2022-12-04&wertetyp=QUARTER_HOUR
> ```
> 
> This is used in the CSV export. Funny enough, why is this the business2business API and not business2consumer as the other API calls? thinking

DarwinsBuddy: > I hesitate to remove the loop though: It is tricky to decide if there are no values stored (and never will be) or if there are no values yet. Thus, if we not got through all values up to now, we might get stuck on a day where no values are stored (and never will be). In such a case, it would be guessing if we should skip the day or if we need to wait until the database is populated.

My thoughts on that:

Dates without data can't happen. So it's not reasonable to skip them.
Further more, when dealing with gaps I can think of following possibilities:

Scenario 1:

There was a gap which will never be filled. 
We cannot do anything about that to get the "real" measurement.
We could "assume" a measurement by interpolating (like with an avg or another estimator), but then we're drifting apart from actually integrating the API to guessing the consumption, which I wouldn't recommend, 'cause I'd like to see that as a responsibility of the data provider "Wiener Netze".

Scenario 2:

There was a gap that will be filled in the future.
This depends a lot on Homeassistant's capabilities, which I do not know. Can't we remove the entries from that day and replace them with the newly fetched ones? Or in general: is replacing possible? If not, then we have to go with that what's fetched the day after. Or another solution: The user can decide when the information is queried. if he/she's fine with it, we could fetch it the day after tomorrow to compensate for unintended gaps in exchange for actuality.

What's your thought on that?

custom_components/wnsm/sensor.py
reox: Yes, I think if there are no values, it is actually an API error.
Maybe it is also a good idea to add the [`raise_for_status`](https://requests.readthedocs.io/en/latest/api/#requests.Response.raise_for_status) after this line: https://github.com/reox/WienerNetzeSmartmeter/blob/6b6fab8919b953234a1ba9f3de89b25f4f05f9e8/custom_components/wnsm/api/client.py#L123
Because I think in such a case, the API will also return a 500

custom_components/wnsm/sensor.py
```diff
@@ -166,29 +282,45 @@ async def async_update(self):
...
+            if not self.is_active(zp):
+                _LOGGER.warning("Smartmeter not active...")
+                return
+
+            # Get last sum and last date from statistics
+            # From
+            # https://github.com/DarkC35/ha_linznetz/blob/904f361e760103f900ad93522a0215d348fc83bb/custom_components/linznetz/sensor.py
+            # Select one entry from the statistics, convert the units
+            # It is crucial to use get_instance here!
+            last_inserted_stat = await get_instance(self.hass).async_add_executor_job(get_last_statistics, self.hass, 1, self._id, True, {"sum", "state"})
```
reox, line 295: Apparently, they made a non-backwards compatible change here. Starting with core 2022.12, you have to specify the types you want to receive from `get_last_statistics`. Furthermore, the statistics return now a datetime object and not str (which is nice)

How should you program for HA? Should you always assume everyone has the latest API? I cannot see anything related to that here: https://developers.home-assistant.io/docs/creating_integration_manifest/#dependencies

custom_components/wnsm/sensor.py
reox: I asked that now: https://community.home-assistant.io/t/sensorentity-insert-historical-statistics-data-but-also-keep-sensor-up-to-date/500880

custom_components/wnsm/sensor.py
cprerovsky: Oh man, just wanted to let you guys know I'm super excited about this project, and specifically this PR. Unfortunately mit Python skills are extremely lacking, so I'll just keep watching for now :D. Thank you guys for putting in the work!

custom_components/wnsm/sensor.py
```diff
@@ -166,29 +284,48 @@ async def async_update(self):
...
+            else:
+                _LOGGER.error(f"unexpected result of get_last_statistics: {last_inserted_stat}")
+                return
+
+            # Collect hourly data
+            await self._import_statistics(smartmeter, start, _sum)
+
             self._available = True
-            self._updatets = datetime.now().strftime("%d.%m.%Y %H:%M:%S")
+            self._updatets = start.strftime("%d.%m.%Y %H:%M:%S")
```
reox, line 328: Actually, using `start` is the wrong date here... the update time is rather the current date, if any data was imported - right?

custom_components/wnsm/api/client.py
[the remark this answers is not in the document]
DarwinsBuddy: nevermind. they changed the order of zaehlpunkte returned by the API, and I didn't specify one, so it defaulted to an abandoned one from an old contract.

To your question: yes! it takes `date_from` and regardgless of the provided `datew_to` data is cut off after 24 hours.

custom_components/wnsm/sensor.py
DarwinsBuddy: love it. hopefully we'll get an answer to that

custom_components/wnsm/api/client.py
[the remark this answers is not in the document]
reox: haha okay, yet another thing for the API thread I guess 😅 
Then I'll leave it as it is now, that you only specify the start date.

custom_components/wnsm/sensor.py
DarwinsBuddy: As it seems home assistant doesn't use semantic versioning, but rather some kind of automatic release every month and patch releases on-demand.

Usually patch releases shouldn't be breaking, but I guess since they do not use semantic versioning, they may lack this kind of feature.

Anyway, I think following the principal of developing against the current home assistant version seems to be the best approach. better to support a version which everybody can get by updating, rather than assuming stuff.

I created an issue to provide (at least for us) a way to track releases. Currently the HEAD of main will be released on push. We can switch to a proper release management to maybe introduce later on tags to support also older versions. (or maybe I find out if there is a way of providing the compatibility information within the manifest)

https://github.com/DarwinsBuddy/WienerNetzeSmartmeter/issues/27

custom_components/wnsm/sensor.py
DarwinsBuddy: correct. I missed that. it should show the timestamp of the last time this sensor was updated.

reox: I reverted that change!

reox:
closing this one in favor of #35

PR #30, opened 2022-12-13
Made Readme config example accesible

TheRealVira:
It wasn't accessible before.


DarwinsBuddy:
Wow thanks that slipped through. ❤️

PR #35, opened 2022-12-31
[WIP] Importing hourly consumption

DarwinsBuddy:
@reox I moved your branch and created a new PR, so we can follow up in the base repo. You should be able to commit.


h0jeZvgoxFepBQ2C:
Does this PR enable the hourly consumption data? right now I only get the daily data imported? Would be wonderful if this could get merged and released soon!

DarwinsBuddy:
> Does this PR enable the hourly consumption data? right now I only get the daily data imported? Would be wonderful if this could get merged and released soon!

@reox is currently working on it. I guess this will take still a bit of time. I guess everyone of us is eager to see this work, but since probably everybody has a job and this is a side project, we'll have to wait still a little longer.

reox:
One issue, I cannot solve right now, is that this does not work as homeassistant intents to...
I.e., the sensor has no state and thus many things do not work as expected.
It seems like this is nothing I can solve from the integration side, but would need some development in HA core as well...

Regardless, I use it now for over a month as a test ans if you accept these limitations, it works quite well.
Still, there are some kinks I want to iron out. ETA is maybe mid feb 😅

h0jeZvgoxFepBQ2C:
Just that I understand it, I guess the problem that we need to import old data (last 24h) into home assistant and there is no good way I guess, since HA usually just saves the last sensor state (and not old / multiple data entries)?

reox:
There are ways to import it into the statistics data directly - you can find a lot of links in #8

But the "official" way (idk if it is because there is only few docs and all I know comes from the HA discourse) requires me to never ever set a state on the sensor directly. 
But that on the other hand confuses many other tools, such as ui cards - because they rely on live data and states on the sensors.
For my purposes that is sufficient, but as I said, there are still a few things I would like to change before merging.

DarwinsBuddy:
@reox I observed, that you had already tried out a dependency of home assistant (`recorder`) which could solve this issue.
if I understood it corrrectly, this would mean that we can't update the sensor periodically and thus it won't be set to active or available in home assistant.

In order to compensate for this inability, maybe we can create two sensors:

One that is merely for representing the ever increasing amount of kWh consumed (which we have right now), not bearing the responsibility for being accurate when it comes to reporting (w.r.t. day specific measurements, cause they can appear on the API way later than they really happened)

Another one that is not active, but bears the information accurately reflecting the real consumption.

Only thing left for clarification is: Is the latter sensor (the one with the accurate historical data) of any use in the energy dashboard? If not, then that is no option unfortunately.

reox:
because of the thousands of errors, I disabled the integration in HA - but enabled it today with the fixes and it fetched the last days without any problem! So the recorder is also somewhat resistant to outages and disabling of the integration.

DarwinsBuddy:
these are absolutely amazing news.
can't wait to see this hit main 🤩

custom_components/wnsm/sensor.py
```diff
@@ -218,57 +296,64 @@ async def async_update(self):
...
+                return
+
+        else:
+            _LOGGER.error(f"unexpected result of get_last_statistics: {last_inserted_stat}")
+            return
+
         try:
             smartmeter = Smartmeter(self.username, self.password)
             await self.hass.async_add_executor_job(smartmeter.login)
             zaehlpunkt = await self.get_zaehlpunkt(smartmeter)
```
reox, line 346: Isn't the Zählpunkt number not stored in the class itself? But we need to fetch this here to check if the ZP is active, right?

DarwinsBuddy: originally, yes.

We could skip that or even better infer the status from the sensor from the call `get_consumption` within your `async_import_statistics` method.

reox: do you know by chance what the reply of the API would be in `get_consumption` if the Zählpunkt is not active? :D

MatthiasRei:
First of all, great job!

@reox I can support you guys with test data. I have 2 accounts, with interesting constellations: 

- Account 1: 1x activated and 1x deactivated "TAGSTROM"
- Account 2: 1x "TAGSTROM" and 1x "BEZUG" (PV)

Let me know what exact data you need :)

DarwinsBuddy:
Hi @MatthiasRei 
I think the scope of this PR is already quite large, let's move this to another task to get onto it, after this here is merged and rolled out:
https://github.com/DarwinsBuddy/WienerNetzeSmartmeter/issues/51

DarwinsBuddy:
@reox how's your feeling about your PR. Do  you think we could merge it anytime soon? (No pressure intended. :blush: )

reox:
So... I tested this since November and it is still working 💪

The things that might need attention / should be tested:
- [ ] What if no optIn was given at time x and later is was granted - will the data be queried? I guess so, because we would at every day check the last 48h and at some point it will start gathering data.
- [ ] what happens at summer/wintertime change? Either someone queries the data from the 27th march 2022 and 30th October 2022 or we wait until the 26th of march ;)
- [ ] to less strain the API, we probably do not need to login every time, but can keep the token, right? But that is probably a general thing

Other than that, I think it would be nice if more people would test it, especially those with multiple zählpunkte!

reox:
There is one thing though I noticed, which is especially visible when there is only the "base load", i.e., I'm not at home:
while the energy dashboard shows everthing correct
image

The widget I use does something funny:
image

the code for the widget is:
```
chart_type: bar
period: hour
days_to_show: 2
type: statistics-graph
entities:
  - sensor.at0010000000000000xxx
stat_types:
  - state
```
I have no idea if this could be prevented by importing the data in a different way or if the widget is just broken.

jozefKruszynski:
> So... I tested this since November and it is still working 💪
> 
> The things that might need attention / should be tested:
> 
> * [ ]  What if no optIn was given at time x and later is was granted - will the data be queried? I guess so, because we would at every day check the last 48h and at some point it will start gathering data.
> * [ ]  what happens at summer/wintertime change? Either someone queries the data from the 27th march 2022 and 30th October 2022 or we wait until the 26th of march ;)
> * [ ]  to less strain the API, we probably do not need to login every time, but can keep the token, right? But that is probably a general thing
> 
> Other than that, I think it would be nice if more people would test it, especially those with multiple zählpunkte!

Hi, I don't have multiple Zählpunkte, but I would like to test this version. How would I go about switching the integration?

fghaas:
I have been testing this since early January, and it's working beautifully for me too, with one exception: while my energy dashboard shows my energy _consumption_ correctly, my energy _cost_ is always 0. 

Prior to running the integration from this topic branch this worked correctly (using the `Use an entity with current price` option, and simply using an input number to define the price), and I am surely doing something wrong, but that's been the only pain point. Other than that, this is looking great from my point of view. Thanks for putting this together!

reox:
> I have been testing this since early January, and it's working beautifully for me too, with one exception: while my energy dashboard shows my energy consumption correctly, my energy cost is always 0.

Yes - this is a major drawback of this branch. The issue is that I simply import the historical statistical data only, while the original implementation updates the energy counter (however with the wrong time of consumption).
Due to lack of functions in HA, nothing else seems to be an option. At least no one in the HA discourse had another idea :/

One idea was to have two energy meters, one that has the correct statistics and one that has the correct meter reading. But I have no clue (yet) how that could be implemented.

jozefKruszynski:
Hi,
I removed the HACS version of this integration and manually installed this branch this morning, how long after setup should I actually see the zählpunkt sensor available in HA? I have tried both setup via UI and via yaml config, and either way I see no sensor being created.

reox:
It should be there immediatetly after logging in. Maybe you still have the old sensor entity?

jozefKruszynski:
> It should be there immediatetly after logging in. Maybe you still have the old sensor entity?

I removed it from the statistics section now, and the new one has appeared. Apologies for the confusion.

reox:
>  Another question, it looks like it pulls the last 48 hours of data, is it possible to get more historic data?

yes, I had this implemented before but we opted to remove it because we are not sure how much we can strain the API... The reason is that we can only query 24h at once and would need to iterate over all days. However in principle it works by changing this line: https://github.com/DarwinsBuddy/WienerNetzeSmartmeter/pull/35/files#diff-2e430aff5503bd2d059e716dacbce1b7556d9137cb0622a892287ff08ece7e8aR322

In general it would be better to provide a field to upload the CSV you can download from the SmartMeter page and import everything from there.

jozefKruszynski:
> > Another question, it looks like it pulls the last 48 hours of data, is it possible to get more historic data?
> 
> yes, I had this implemented before but we opted to remove it because we are not sure how much we can strain the API... The reason is that we can only query 24h at once and would need to iterate over all days. However in principle it works by changing this line: https://github.com/DarwinsBuddy/WienerNetzeSmartmeter/pull/35/files#diff-2e430aff5503bd2d059e716dacbce1b7556d9137cb0622a892287ff08ece7e8aR322
> 
> In general it would be better to provide a field to upload the CSV you can download from the SmartMeter page and import everything from there.

In the meantime I found this line and changed it to 1440 hours which works out around when my smart meter was activated. There were a few log lines about no granular measurements active, and then everything else was updated.

If I now change the back to 48 hours, and restart home assistant, it appears that there are errors in the logs regarding float object has no attribute replace:

```
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 507, in _async_add_entity
    await entity.async_device_update(warning=False)
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 732, in async_device_update
    await task
  File "/config/custom_components/wnsm/sensor.py", line 334, in async_update
    delta_t = datetime.now(timezone.utc).replace(microsecond=0) - start.replace(microsecond=0)
AttributeError: 'float' object has no attribute 'replace'
```

reox:
Interesting, what is your HA version? Looks like you got back a float instead of datetime. There was a change last year in the API where they changed it from float to datetime afair

jozefKruszynski:
> Interesting, what is your HA version? Looks like you got back a float instead of datetime. There was a change last year in the API where they changed it from float to datetime afair

I just updated to 2023.3.0 last night, but I usually update within a day or 2 of a new release in general.

reox:
okay thats weird... can you maybe turn up debugging for the integration and post the lines that start with `Last inserted stat` - i.e. the output of https://github.com/DarwinsBuddy/WienerNetzeSmartmeter/pull/35/files#diff-2e430aff5503bd2d059e716dacbce1b7556d9137cb0622a892287ff08ece7e8aR314

just add
```
logger:
  logs:
    custom_components.wnsm: debug
```
to the `configuration.yaml`.

reox:
Well - I updated to 2023.3.0 as well and are greeted with this:

Logger: homeassistant.components.sensor
Source: components/recorder/util.py:207
Integration: Sensor (documentation, issues)
First occurred: 10:58:48 (1 occurrences)
Last logged: 10:58:48

```
wnsm: Error on device update!
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1964, in _exec_single_context
    self.dialect.do_execute(
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 747, in do_execute
    cursor.execute(statement, parameters)
sqlite3.OperationalError: no such column: statistics.start_ts
```

The above exception was the direct cause of the following exception:

```
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 507, in _async_add_entity
    await entity.async_device_update(warning=False)
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 732, in async_device_update
    await task
  File "/config/custom_components/wnsm/sensor.py", line 303, in async_update
    last_inserted_stat = await get_instance(
  File "/usr/local/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/src/homeassistant/homeassistant/components/recorder/statistics.py", line 1900, in get_last_statistics
    return _get_last_statistics(
  File "/usr/src/homeassistant/homeassistant/components/recorder/statistics.py", line 1872, in _get_last_statistics
    stats = cast(Sequence[Row], execute_stmt_lambda_element(session, stmt))
  File "/usr/src/homeassistant/homeassistant/components/recorder/util.py", line 207, in execute_stmt_lambda_element
    executed = session.execute(stmt)
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/orm/session.py", line 2229, in execute
    return self._execute_internal(
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/orm/session.py", line 2124, in _execute_internal
    result: Result[Any] = compile_state_cls.orm_execute_statement(
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/orm/context.py", line 253, in orm_execute_statement
    result = conn.execute(
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1414, in execute
    return meth(
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/sql/lambdas.py", line 600, in _execute_on_connection
    return connection._execute_clauseelement(
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1638, in _execute_clauseelement
    ret = self._execute_context(
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1842, in _execute_context
    return self._exec_single_context(
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1983, in _exec_single_context
    self._handle_dbapi_exception(
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 2325, in _handle_dbapi_exception
    raise sqlalchemy_exception.with_traceback(exc_info[2]) from e
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1964, in _exec_single_context
    self.dialect.do_execute(
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 747, in do_execute
    cursor.execute(statement, parameters)
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such column: statistics.start_ts
[SQL: SELECT statistics.metadata_id, statistics.start_ts, statistics.mean, statistics.min, statistics.max, statistics.last_reset_ts, statistics.state, statistics.sum 
FROM statistics 
WHERE statistics.metadata_id = ? ORDER BY statistics.metadata_id, statistics.start_ts DESC
 LIMIT ? OFFSET ?]
[parameters: (47, 1, 0)]
(Background on this error at: https://sqlalche.me/e/20/e3q8)
```

jozefKruszynski:
> okay thats weird... can you maybe turn up debugging for the integration and post the lines that start with `Last inserted stat` - i.e. the output of https://github.com/DarwinsBuddy/WienerNetzeSmartmeter/pull/35/files#diff-2e430aff5503bd2d059e716dacbce1b7556d9137cb0622a892287ff08ece7e8aR314
> 
> just add
> 
> ```
> logger:
>   logs:
>     custom_components.wnsm: debug
> ```
> 
> to the `configuration.yaml`.
> 
> edit: I remember, the change back then was that a datetime was returned instead a str, maybe they changed it now to float?!


2023-03-02 11:10:25.200 DEBUG (MainThread) [custom_components.wnsm.sensor] Last inserted stat: defaultdict(<class 'list'>, {'sensor.at0010000000000000001000015392393': [{'start': 1677708000.0, 'end': 1677711600.0, 'state': 0.181, 'sum': 251.276}]})

reox:
Thanks!
So indeed, you get a float now... oO

jozefKruszynski:
> I just pushed a workaround and also asked in the HA community: https://community.home-assistant.io/t/get-last-statistics-returned-str-then-datetime-now-float/542311

Seems to fix the logs at least.

jozefKruszynski:
> I just pushed a workaround and also asked in the HA community: https://community.home-assistant.io/t/get-last-statistics-returned-str-then-datetime-now-float/542311

Looks like there was indeed in the changeling a breaking change listed to statistics that relates to using timestamp columns

https://www.home-assistant.io/changelogs/core-2023.3

https://github.com/home-assistant/core/pull/87321

reox:
> Looks like there was indeed in the changeling a breaking change listed to statistics that relates to using timestamp columns

Thanks for digging this up. It's funny that the "only" breaking change is the unavailability during migration... ¯\_(ツ)_/¯

jozefKruszynski:
> > I have been testing this since early January, and it's working beautifully for me too, with one exception: while my energy dashboard shows my energy consumption correctly, my energy cost is always 0.
> 
> Yes - this is a major drawback of this branch. The issue is that I simply import the historical statistical data only, while the original implementation updates the energy counter (however with the wrong time of consumption). Due to lack of functions in HA, nothing else seems to be an option. At least no one in the HA discourse had another idea :/
> 
> One idea was to have two energy meters, one that has the correct statistics and one that has the correct meter reading. But I have no clue (yet) how that could be implemented.

There should be an associated statistics_meta entry for the cost sensor that gets created along with the consumption sensor right?

Would it be possible to write the statistics entries for the cost sensor at the same time as writing the statistics for the consumption sensor?

I noticed that with this branch, the cost sensor does not get created in the statistics_meta table, however, there is a sensor that is created with the same name as the consumption sensor suffixed with "_cost"

Would it then be feasible to take an input_number sensor for example and use this as the basis for the statistics entries for the cost sensor?

I think that in order for this to work, you would also first need to create the meta entry, but perhaps it might do the trick...

jozefKruszynski:
> I do not know the internals of HA that much, but I would guesstimate that writing statistics does not help, as statistics are not states.
> One idea would be to have two sensors, one that has the hourly consumption stats and one that has the meter reading + costs.
> Unfortunately, both cannot be mixed.
> I don't even have any costs entity, but maybe that is because I haven't configured any costs?

The costs entity is only visible in the devtools -> statistics tab.

reox:
ah, you have to configure costs for the sensor first. I had checked "do not track costs".

DarwinsBuddy:
@reox I've to admit that, I haven't installed the PR on my system. afaik it replaces the sensor by its historical counterpart.
Are there any drawbacks to it? Afaik that we cannot trigger any automations from it, right?
If that's the case I'd create a second sensor working as used to be. So the integration would create two sensors per zählpunkt. One statistics and one "live" one (which is updated presumably ever day).

Is this correct and also a suitable solution iyo?

reox:
> Is this correct and also a suitable solution iyo?

yes! I started to implement this, but got lost in setting up the devcontainer.. apparently, the one that is configured right now is deprecated and shall not be used.

So, my plan was to add another parameter to `__init__` of the Sensor to switch between a statistics sensor and a counting sensor.
But I'm not sure if this is the right way to go...

reox:
Now I feel stupid.. On closer thought on the summer/wintertime issue: it is no issue at all. UTC is monotonic and does not jump around. Thus, I'm certain that the things get imported correctly and just the UI has to display them properly.

DarwinsBuddy:
Sorry for the late reply.
Unfortunately I was involved in an accident and am slowly recovering.

I'll try to split this up into two separate sensors then.
I also have to loosen up the coverage thresholds, so they are not blocking us atm.

DarwinsBuddy:
> > Is this correct and also a suitable solution iyo?
> 
> yes! I started to implement this, but got lost in setting up the devcontainer.. apparently, the one that is configured right now is deprecated and shall not be used.
> 
> So, my plan was to add another parameter to `__init__` of the Sensor to switch between a statistics sensor and a counting sensor. But I'm not sure if this is the right way to go...

should we remove it again @TheRealVira ?

salzrat:
I would really love the hourly consumption - does it make sense to try the statistics branch yet, or should I wait a bit more?

reox:
> I would really love the hourly consumption - does it make sense to try the statistics branch yet, or should I wait a bit more?

sure! If you are starting from scratch anyways, maybe test #89 directly :)

salzrat:
No, I'm not starting from scratch, I have the main branch running for several months already. Above are some mentions of things that still need to be done. Will they impact the experience if I try it now or should I be good to go?

PR #101, opened 2023-04-08
Initial Import of Historical Data

### LINKED ISSUE #26: Add csv import capability and hook it up to init sensor (or config flow)
https://github.com/DarwinsBuddy/WienerNetzeSmartmeter/pull/20#discussion_r1038999796

IDEA:

Make use of the Wiener Netze csv export endpoint
Implement an import of csv into statistics module of HA
hook it up to config flow (so it can be triggered on demand)
call it on initialization of the integration (when setting up the sensor)

Thanks @reox for this idea!

reox:
yet another thing to test ;)

Implemented the fetching of the historical data and import it on initial sensor setup.

This resolves #26 - at least partially, because it grabs the data directly instead of asking


reox:
@DarwinsBuddy I do not understand why the tests are failing here. The response should be mocked including the expire times, and thus also on calling `login()` also the expire variables should be set - no?

DarwinsBuddy:
> @DarwinsBuddy I do not understand why the tests are failing here. The response should be mocked including the expire times, and thus also on calling `login()` also the expire variables should be set - no?

My bad. I made a mistake setting up that test case

DarwinsBuddy:
I launched it, got a warning stating that the updating of the sensor took > 10 seconds. Maybe 3 years is a bit far in the past for one shot.

Afterwards I was not able to see any historical data in the sensor. Where should I be able to see the pulled data?

custom_components/wnsm/api/client.py
```diff
@@ -84,30 +92,58 @@ def login(self):
                 f"Could not obtain access token: {result.content}"
             )
 
-        self._access_token = result.json()["access_token"]
-        self._refresh_token = result.json()["refresh_token"] # TODO: use this to refresh the token of this session instead of re-login. may be nicer for the API
-        self._api_gateway_token = self._get_api_key(self._access_token)
+        logger.debug(f"LOGIN RESULT: {result.json()}")
+        res_json = result.json()
+        if res_json['token_type'] != 'Bearer':
+            raise SmartmeterConnectionError(f'Bearer token required, but got {res_json["token_type"]!r}')
```
DarwinsBuddy, line 98: Please add a test for this

custom_components/wnsm/api/client.py
```diff
@@ -84,30 +92,58 @@ def login(self):
...
+
+        self._api_gateway_token, self._api_gateway_b2b_token = self._get_api_key(self._access_token)
+        return self
 
     def _get_api_key(self, token):
+        key_b2c = None
+        key_b2b = None
+
+        if datetime.now() >= self._access_token_expiration:
+            raise SmartmeterConnectionError("Access Token is not valid anymore, please re-log!")
```
DarwinsBuddy, line 117: Please add a test for this

custom_components/wnsm/api/client.py
```diff
@@ -122,20 +158,35 @@ def _call_api(
         query=None,
         return_response=False,
         timeout=60.0,
+        extra_headers=None,
     ):
+        if datetime.now() >= self._access_token_expiration:
+            raise SmartmeterConnectionError("Access Token is not valid anymore, please re-log!")
```
DarwinsBuddy, line 164: Please add a test for this

custom_components/wnsm/api/client.py
```diff
@@ -122,20 +158,35 @@ def _call_api(
...
         headers = {
             "Authorization": f"Bearer {self._access_token}",
-            "X-Gateway-APIKey": self._api_gateway_token,
         }
 
+        # For API calls to B2C or B2B, we need to add the Gateway-APIKey:
+        if base_url == const.API_URL:
+            headers['X-Gateway-APIKey'] = self._api_gateway_token
+        elif base_url == const.API_URL_B2B:
+            headers['X-Gateway-APIKey'] = self._api_gateway_b2b_token
```
DarwinsBuddy, line 183: Please add a test for this

custom_components/wnsm/api/client.py
```diff
@@ -122,20 +158,35 @@ def _call_api(
...
+
+        if extra_headers:
+            headers.update(extra_headers)
```
DarwinsBuddy, line 186: Please add a test for this

custom_components/wnsm/api/client.py
```diff
@@ -122,20 +158,35 @@ def _call_api(
...
+
         if data:
+            logger.debug("DATA: %s" % data)
```
DarwinsBuddy, line 189: Please add a test for this
Reconsider logging of this data

custom_components/wnsm/api/client.py
```diff
@@ -194,50 +248,46 @@ def verbrauch_raw(
...
-            zaehlpunkt = self._get_first_zaehlpunkt()
-        endpoint = f"messdaten/zaehlpunkt/{zaehlpunkt}/verbrauch"
-        query = const.build_verbrauchs_args(
-            dateFrom=self._dt_string(day.replace(hour=0, minute=0, second=0))
-        )
-        return self._call_api(endpoint, query=query)
+        # FIXME: Actually, using 00:00:00.000 does not query the beginning of the day!
+        # The problem is, that the time has to specified in UTC and during standard time,
+        # UTC day starts at 23:00:00 and during summer time even on 22:00:00!
+        return self.verbrauch(day.replace(hour=0, minute=0, second=0, microsecond=0), zaehlpunkt, resolution)
```
DarwinsBuddy, line 290: Please add a test for this

custom_components/wnsm/api/client.py
```diff
@@ -306,3 +356,54 @@ def create_ereignis(self, zaehlpunkt, name, date_from, date_to=None):
...
+        valuetype: const.ValueType = const.ValueType.QUARTER_HOUR,
+    ):
+        """
+        Query historical data in a batch
+
+        If no arguments are given, a span of three year is queried (same day as today but from current year - 3).
+        If date_from is not given but date_until, again a three year span is assumed.
+        """
+        if zaehlpunkt is None:
+            zaehlpunkt = self._get_first_zaehlpunkt()
```
DarwinsBuddy, line 374: Please add a test for this

custom_components/wnsm/api/client.py
```diff
@@ -306,3 +356,54 @@ def create_ereignis(self, zaehlpunkt, name, date_from, date_to=None):
...
+
+        if date_until is None:
+            date_until = date.today()
```
DarwinsBuddy, line 377: Please add a test for this

custom_components/wnsm/api/client.py
```diff
@@ -306,3 +356,54 @@ def create_ereignis(self, zaehlpunkt, name, date_from, date_to=None):
...
+
+        if date_from is None:
+            date_from  = date_until.replace(year=date_until.year - 3)
```
DarwinsBuddy, line 380: Please add a test for this

custom_components/wnsm/api/client.py
```diff
@@ -306,3 +356,54 @@ def create_ereignis(self, zaehlpunkt, name, date_from, date_to=None):
...
+
+        query = {
```
DarwinsBuddy, line 382: Please add a test for this

custom_components/wnsm/api/client.py
```diff
@@ -306,3 +356,54 @@ def create_ereignis(self, zaehlpunkt, name, date_from, date_to=None):
...
+            'zaehlpunkt': zaehlpunkt,
+            'datumVon': date_from.strftime('%Y-%m-%d'),
+            'datumBis': date_until.strftime('%Y-%m-%d'),
+            'wertetyp': valuetype.value,
+        }
+
+        extra = {
```
DarwinsBuddy, line 389: Please add a test for this

custom_components/wnsm/api/client.py
```diff
@@ -306,3 +356,54 @@ def create_ereignis(self, zaehlpunkt, name, date_from, date_to=None):
...
+            # For this API Call, requesting json is important!
+            "Accept": "application/json"
+        }
+
+        data = self._call_api('zaehlpunkte/messwerte', base_url=const.API_URL_B2B, query=query, extra_headers=extra)
```
DarwinsBuddy, line 394: Please add a test for this

custom_components/wnsm/api/client.py
```diff
@@ -306,3 +356,54 @@ def create_ereignis(self, zaehlpunkt, name, date_from, date_to=None):
...
+        # Some Sanity Checks...
+        if len(data) != 1 or data[0]['zaehlpunkt'] != zaehlpunkt or len(data[0]['zaehlwerke']) != 1:
+            # TODO: Is it possible to have multiple zaehlwerke in one zaehlpunkt?
+            # I guess so, otherwise it would not be a list...
+            # Probably (my guess), we would see this on the OBIS Code.
+            # The OBIS Code can code for channels, thus we would probably see that there.
+            # Keep that in mind if for someone this fails.
+            logger.debug(f"Returned data: {data}")
+            raise SmartmeterQueryError("Returned data does not match given zaehlpunkt!")
+        obis_code = data[0]['zaehlwerke'][0]['obisCode']
```
DarwinsBuddy, line 405: Please add a test for this

custom_components/wnsm/api/client.py
```diff
@@ -306,3 +356,54 @@ def create_ereignis(self, zaehlpunkt, name, date_from, date_to=None):
...
+        if obis_code[0] != '1':
+            logger.warning(f"The OBIS code of the meter ({obis_code}) reports that this meter does not count electrical energy!")
```
DarwinsBuddy, line 407: Please add a test for this

custom_components/wnsm/api/client.py
```diff
@@ -84,30 +92,58 @@ def login(self):
                 f"Could not obtain access token: {result.content}"
             )
 
-        self._access_token = result.json()["access_token"]
-        self._refresh_token = result.json()["refresh_token"] # TODO: use this to refresh the token of this session instead of re-login. may be nicer for the API
-        self._api_gateway_token = self._get_api_key(self._access_token)
+        logger.debug(f"LOGIN RESULT: {result.json()}")
```
DarwinsBuddy, line 95: Commits introducing logging of sensitive data (even if it's debug) is maybe not a good idea

custom_components/wnsm/api/client.py
```diff
@@ -57,13 +64,14 @@ def login(self):
         except Exception as exception:
             raise SmartmeterConnectionError("Could not login with credentials") from exception
 
+        logger.debug("LOGIN HEADERS: %s" % result.headers)
```
DarwinsBuddy, line 67: Commits introducing logging of sensitive data (even if it's debug) is maybe not a good idea

DarwinsBuddy (changes requested):
Please run

`flake8 custom_components/wnsm --count --select=E9,F63,F7,F82 --show-source --statistics --config tests/setup.cfg`
and
`flake8 custom_components/wnsm --count --exit-zero --max-complexity=10 --statistics --config tests/setup.cfg`

and resolve the issues. (sometimes it pays of to disable them inline ` #noqa: XXX`

reox:
> That is amazing! How did you find out this is a feature of the b2b api?

just watched the webpage with the developer tools :grinning: 

> Maybe 3 years is a bit far in the past for one shot.

When you visit the dataexport in the smartmeter-web, 3 years is the standard setting. Thus I thought it might be OK to do the same. Does it take very long for you? I only have smartmeter entries since November 2023, and thus with any date range it is actually quite fast.

I just thought about an issue with this approach: If for some reason it fails to gather the data, the init procedure will be called again and again. So we need some way of determining that the init was called once and do not start it again.

custom_components/wnsm/api/client.py
reox: Ah yes, that is true. I'll remove them!

custom_components/wnsm/api/client.py
reox: this data thing is actually from the original API. It is only used for the `user/ereignis` endpoint - which I never used.

custom_components/wnsm/api/client.py
```diff
@@ -1,6 +1,5 @@
 """Contains the Smartmeter API Client."""
 import logging
-import pprint
```
DarwinsBuddy: thanks :grimacing: I forgot this

custom_components/wnsm/api/client.py
DarwinsBuddy: ah I see. then let's remove this for good. the less individual user data logging, the better.

DarwinsBuddy:
> > That is amazing! How did you find out this is a feature of the b2b api?
> 
> just watched the webpage with the developer tools grinning
> 
So same technique as I did it. :) but I've not discovered this part yet. Good catch! :heart: 

> > Maybe 3 years is a bit far in the past for one shot.
> 
> When you visit the dataexport in the smartmeter-web, 3 years is the standard setting. Thus I thought it might be OK to do the same. Does it take very long for you? I only have smartmeter entries since November 2023, and thus with any date range it is actually quite fast.
> 
> I just thought about an issue with this approach: If for some reason it fails to gather the data, the init procedure will be called again and again. So we need some way of determining that the init was called once and do not start it again.


hmm... I guess then 3years should be fine. But you have a point there. What we could do though is only try it once at set-up. If it fails we shouldn't trigger it anew. We can implement an option in the config_flow to trigger it manually. I have to look into the details though. so better add an `at most once` logic, create another issue and let's solve this after this monster of a branch is in `main` :smiley:

custom_components/wnsm/api/client.py
reox: This function is only ever used in `live_sensor` as a fallback. The question here is what this function now actually queries. It correctly sets the day to 00:00:00 UTC - but does not check if a timezone is set in the datetime. However, probably what should be done here is to set it to 00:00:00 Local time, i.e. probably CE(S)T.

custom_components/wnsm/api/client.py
reox: it would be interesting to know how the data looks like for others... All these checks were just assumptions I made by looking at my own data. But I only have one zählpunkt and not the longest coverage of values.
Thus, it is a bit hard to test right now :sweat_smile:

DarwinsBuddy:
@reox unless you have any concerns. I'd release this feature. It's downwards compatible and runs smoothly already.

Regarding 
 
> > When you visit the dataexport in the smartmeter-web, 3 years is the standard setting. Thus I thought it might be OK to do the same. Does it take very long for you? I only have smartmeter entries since November 2023, and thus with any date range it is actually quite fast.
> > I just thought about an issue with this approach: If for some reason it fails to gather the data, the init procedure will be called again and again. So we need some way of determining that the init was called once and do not start it again.
> 
> hmm... I guess then 3years should be fine. But you have a point there. What we could do though is only try it once at set-up. If it fails we shouldn't trigger it anew. We can implement an option in the config_flow to trigger it manually. I have to look into the details though. so better add an `at most once` logic, create another issue and let's solve this after this monster of a branch is in `main` smiley

I'd create another issue, when I've got your Okay. :)

reox:
@DarwinsBuddy  it runs also smoothly on my side! Except some missing test cases, I think it is good to go :)

custom_components/wnsm/api/client.py
DarwinsBuddy: see #113

custom_components/wnsm/api/client.py
DarwinsBuddy: see #113

custom_components/wnsm/api/client.py
DarwinsBuddy: I guess the WienerNetzeAPI is not caring for timezones, thus it's assuming Europe/Vienna as the default timezone when submitting UTC. But that's just a wild guess.

Also see  #113

custom_components/wnsm/api/client.py
DarwinsBuddy: see #113

custom_components/wnsm/api/client.py
DarwinsBuddy: see #113

custom_components/wnsm/api/client.py
DarwinsBuddy: see #113

custom_components/wnsm/api/client.py
DarwinsBuddy: see #113

custom_components/wnsm/api/client.py
DarwinsBuddy: see #113

custom_components/wnsm/api/client.py
DarwinsBuddy: see #113

custom_components/wnsm/api/client.py
DarwinsBuddy: see #113

DarwinsBuddy (approved):
lgtm

Issue #119, opened 2023-04-25
Entität nicht definiert / not defined


GeraldPape:
Added the attribute to the energy tab, receiving the following error:
![grafik](https://user-images.githubusercontent.com/72608781/234248639-f517b675-5b9b-49a1-87d6-4836edaf3635.png)

for reference:
in the smart meter portal the "newest" data for my smart meter is from 15 days ago.
My data history starts on the 24. February 2023

reox:
>  in the smart meter portal the "newest" data for my smart meter is from 15 days ago.

If there is no data in the smartmeter-web, it cannot be imported into HA. We use the same API as the smartmeter-web does...

Other than that, this warning should not matter and statistics should still be imported.

PR #123, opened 2023-04-30
Change statistics_sensor to measurement and set it to 0 kWh at the end

DarwinsBuddy:
After reading https://github.com/DarwinsBuddy/WienerNetzeSmartmeter/issues/119 I wondered what the difference in set up is for live vs. statistics sensor.

I remember, that @reox had troubles setting the state of the statistics entity of a smart meter to something not `None`.
This presumably leads to the reported warning.

In order to fix this I tried out setting the statistics sensor to `measurement` (since it's not quite `total_increasing`) and set its value to `0 kWh`, s.t. it has an entry in the entity registry of home assistant (which was not the case prior to that).

The warning in the energy dashboard is gone.

@reox Can you potentially confirm that I didn't break anything else with this change?

### DIFF
--- a/custom_components/wnsm/base_sensor.py
+++ b/custom_components/wnsm/base_sensor.py
@@ -48,7 +48,6 @@ class BaseSensor(SensorEntity, ABC):
         self._attr_name = zaehlpunkt
         self._attr_icon = self._icon()
         self._attr_device_class = SensorDeviceClass.ENERGY
-        self._attr_state_class = SensorStateClass.TOTAL_INCREASING
         self._attr_native_unit_of_measurement = UnitOfEnergy.KILO_WATT_HOUR
         self._attr_unit_of_measurement = UnitOfEnergy.KILO_WATT_HOUR
 
@@ -57,6 +56,10 @@ class BaseSensor(SensorEntity, ABC):
         self._state: int | str | None = None
         self._available: bool = True
         self._updatets: str | None = None
+
+    @property
+    def _attr_state_class(self):
+        return SensorStateClass.TOTAL_INCREASING
 
     @property
     def _id(self):
--- a/custom_components/wnsm/statistics_sensor.py
+++ b/custom_components/wnsm/statistics_sensor.py
@@ -1,7 +1,7 @@
 import logging
 
 from homeassistant.components.recorder import get_instance
-from homeassistant.components.sensor import SensorEntity
+from homeassistant.components.sensor import SensorEntity, SensorStateClass
 
 from .api import Smartmeter
 from .base_sensor import BaseSensor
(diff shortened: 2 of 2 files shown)


reox:
interesting! maybe that works, but I have to test that... The interesting thing is that we also import the state statistics, but maybe they are not overwritten if the state is zero? Eventually I know too little about the HA internals :D

DarwinsBuddy:
> interesting! maybe that works, but I have to test that... The interesting thing is that we also import the state statistics, but maybe they are not overwritten if the state is zero? Eventually I know too little about the HA internals :D

I can't answer that, but if the change does not break anything. (i.e. the statistics sensor keeps importing data successively), the warning would be gone and thus potential future issues :upside_down_face:

reox:
Nothing to complain about the code, I switched to that branch now and let it run for a few days. But I'll guess it's fine!

reox:
I get this error now:

Dieser Fehler wurde von einer benutzerdefinierten Integration verursacht

Logger: homeassistant.helpers.entity
Source: custom_components/wnsm/statistics_sensor.py:81
Integration: WienerNetzeSmartmeter (documentation, issues)
First occurred: 01:57:11 (6 occurrences)
Last logged: 06:57:11

```
Update for sensor.at0010000000000000001000015166154_statistics fails
File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 554, in async_update_ha_state
await self.async_device_update()
File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 784, in async_device_update
await coro
File "/config/custom_components/wnsm/statistics_sensor.py", line 81, in async_update
_sum = Decimal(last_inserted_stat[self._id][0]["sum"])
TypeError: conversion from NoneType to Decimal is not supported
```

Not sure how the sum got None all of a sudden? Maybe the switch of the sensor type caused that?
Looks like this fix is not backwards compatible with old measurements - or @DarwinsBuddy did that worked for you and my setup is just a bit weird?

reox:
@DarwinsBuddy I removed the intgration and deleted the statistics from the database to start from scratch. In that case, it works and the `_statistics` sensor shows 0kWh.
Do you understand what's going on when there is already data and the sensor is switched? Otherwise the only chance would be to remove the statistics and start all over. With the import feature that would not be a great loss though.

DarwinsBuddy:
I think removing and regenerating the history is definitely an option. though I am not sure how to achieve this after the upgrade. I know of that one service built in to hassio which is easy to call at least: purge entities.

but I'd have to research if we could call it from within our custom component.
if not we can directly use the recorder I guess?

still what's left is too check somehow when it's been updated to the next version and do it only then.

reox:
I got the error again - this time on the freshly imported data.
I guess setting the state to anything else than None is a bad idea :(
```
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 554, in async_update_ha_state
    await self.async_device_update()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 784, in async_device_update
    await coro
  File "/config/custom_components/wnsm/statistics_sensor.py", line 81, in async_update
    _sum = Decimal(last_inserted_stat[self._id][0]["sum"])
TypeError: conversion from NoneType to Decimal is not supported
```

PR #144, opened 2023-05-22
Check if historical data was returned

### LINKED ISSUE #143: v1.1.0 --> statistics_sensor --> list index out of range
Hi @DarwinsBuddy 

I go a new smart Meter with customer interface and I get now the below message on the statistic sensor.
Not sure if the change is related to I guess not.
I use as well 15 mins value in the Smart meter Portal.

Error message is given as.

Logger: homeassistant.components.sensor
Source: custom_components/wnsm/statistics_sensor.py:179
Integration: Sensor ([documentation](https://www.home-assistant.io/integrations/sensor), [issues](https://github.com/home-assistant/home-assistant/issues?q=is%3Aissue+is%3Aopen+label%3A%22integration%3A+sensor%22))
First occurred: 23:43:55 (1 occurrences)
Last logged: 23:43:55

wnsm: Error on device update!
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 521, in _async_add_entity
    await entity.async_device_update(warning=False)
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 779, in async_device_update
    await self.async_update()
  File "/config/custom_components/wnsm/statistics_sensor.py", line 126, in async_update
    await self._import_historical_data(smartmeter)
  File "/config/custom_components/wnsm/statistics_sensor.py", line 179, in _import_historical_data
    _LOGGER.debug(f"Importing statistics from {statistics[0]} to {statistics[-1]}")
IndexError: list index out of range

reox:
resolves #143

There is just one question now: In this case, the historical data import will be triggered every time until there is any data.
@DarwinsBuddy Should we write an empty statistics here to prevent the import the next time?

### DIFF
--- a/custom_components/wnsm/statistics_sensor.py
+++ b/custom_components/wnsm/statistics_sensor.py
@@ -175,6 +175,11 @@ class StatisticsSensor(BaseSensor, SensorEntity):
         for ts, usage in sorted(dates.items(), key=itemgetter(0)):
             total_usage += usage
             statistics.append(StatisticData(start=ts, sum=total_usage, state=usage))
+        
+        if len(statistics) == 0:
+             # No data is returned, possibly the smart meter is too new or not active yet...
+            _LOGGER.info("No historical data available for smart meter!")
+            return
 
         _LOGGER.debug(f"Importing statistics from {statistics[0]} to {statistics[-1]}")
         async_import_statistics(self.hass, metadata, statistics)


DarwinsBuddy:
Empty one might be the safe call. Only prerequisite for it should be that we definitely know that there is no data to import, and it's not a flaky response

reox:
>  Only prerequisite for it should be that we definitely know that there is no data to import, and it's not a flaky response

I guess this is very hard to test... Or can we maybe query somehow when the smartmeter was activated? Then it would be relatively easy to decide if there is data to be expected.

flo071 (changes requested):
Works like a charm, when it comes to the point that there are no errors thrown anymore, but we do not get the info, change it to `_LOGGER.debug("")` and it would be fine.

DarwinsBuddy (approved):
lgtm

PR #206, opened 2023-12-24
Bugfix/verbrauch

reox:
go back to `verbrauch` for 24h hourly data.
Clean up the API calls a bit.

### DIFF
--- a/custom_components/wnsm/statistics_sensor.py
+++ b/custom_components/wnsm/statistics_sensor.py
@@ -214,21 +214,29 @@ class StatisticsSensor(BaseSensor, SensorEntity):
             last_ts = start
             start += timedelta(hours=24)  # Next batch. Setting this here should avoid endless loops
 
-            if 'values' not in consumption:
-                _LOGGER.error(f"No values in API response! This likely indicates an API error. Original response: {consumption}")
-                return
 
             # Check if this batch of data is valid and contains hourly statistics:
             if not consumption.get('optIn'):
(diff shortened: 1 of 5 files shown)


custom_components/wnsm/statistics_sensor.py
```diff
@@ -205,35 +205,43 @@ async def _import_statistics(self, smartmeter: Smartmeter, start: datetime, tota
...
+                _LOGGER.warning(f"Data starting at {start} does not contain granular data! Opt-in was not set back then. Skipping.")
                 continue
 
             # Can actually check, if the whole batch can be skipped.
             if consumption.get('consumptionMinimum') == 0 and consumption.get('consumptionMaximum') == 0:
-                _LOGGER.debug("Batch of data does not contain any consumption, skipping")
+                _LOGGER.debug(f"Batch of data starting at {start} does not contain any consumption, skipping")
                 continue
 
+            if 'values' not in consumption:
```
reox, line 226: Maybe this change here is not required: The `verbrauch` endpoint should always return values, but if no stuff was recorded or is not available, these values are zero. Thus it looks like that in contrary to `verbrauchRaw`, no `values` is indeed indicating an error.
What to do?

reox:
I remove the `dateTo` from the function, because it is not required (anymore?). But where is that still defined in the test?

tixoman (commented):
passt und funktioniert

tixoman (approved):

DarwinsBuddy (approved):
lgtm

PR #221, opened 2024-03-03
fix negative consumption and add explaination

### LINKED ISSUE #219: Energy Dashboard shows feeding without actually feeding in place
Hi,
as I had some issues the last days - missing data points and if they are shown it was wrong data, I deleted the integration and historical data and started from scratch.
Everything showed as correct, but now I realized that it shows that I seemed to feed around 12kwh today. But I do not feed energy at all.
energy
I found this entry - it is around the time when I setup the integration
energy_stat
It seems there is some calculation wrong imho.
TIA

reox:
This should resolve #219


DarwinsBuddy (approved):

DarwinsBuddy:
was busy the last weeks, sorry for the long line

PR #245, opened 2024-07-16
Update constants.py to include AnlageType WAERMEPUMPE

Skaldhor:
Update constants.py to include AnlageType WAERMEPUMPE


DarwinsBuddy (approved):
lgtm

PR #254, opened 2024-08-30
Fix same zaehlpunkt in multiple contracts

### LINKED ISSUE #253: Conflicting entries/data when having same zaehlpunkt in multiple contracts
I have switched power provider and also made a change to the contract partner. So I received a new contract number for the same zaehlpunkt. When having both the new and old contract in the smart meter, both zaehlpunkte show up, one as active one as inactive. When adding to home assistant, it tries to add the same zaehlpunkt twice and then doesn't update the data properly, since it sometimes updates the data with the one which is not active.

Here a picture of the smart meter portal to get a better understanding of the complication:
grafik

I implemented the fix. It now checks if the zaehlpunkt is active when adding it and when updating it. PR will follow.

tschoerk:
Fixes #253


DarwinsBuddy (approved):
under the assumption that `isActive` is always present this seems to be a safe fix

PR #261, opened 2024-09-20
Bugfix/multiple zps config yaml

DarwinsBuddy:


custom_components/wnsm/api/client.py
tschoerk, line 1: This is from PR #262, should probably be removed from this PR.

custom_components/wnsm/base_sensor.py
tschoerk, line 1: This is from PR #262, should probably be removed from this PR.

custom_components/wnsm/config_flow.py
tschoerk, line 1: This is from PR #262, should probably be removed from this PR.

custom_components/wnsm/sensor.py
tschoerk, line 1: Elegant solution, but I get the following error when trying to install it via configuration.yaml. Something with the way homeassistant handles the configuration. I don't know too much about this, maybe we can figure it out.

`Invalid config for 'sensor' from integration 'wnsm' at configuration.yaml, line 28: value should be a string for dictionary value 'device_id', got ['AT0010000000000000001000000000001', 'AT0010000000000000001000000000002', 'AT0010000000000000001000000000003'], please check the docs at https://github.com/DarwinsBuddy/WienerNetzeSmartmeter`

example/secrets.yaml
```diff
@@ -0,0 +1,6 @@
+vsm_username: example@mail.com
+vsm_password: 12345
+vsm_zaehlpunktnummer:
+  - AT0010000000000000001000000000001
+  - AT0010000000000000001000000000002
+  - AT0010000000000000001000000000003
```
tschoerk, line 6: Elegant solution, but I get the following error when trying to install it via configuration.yaml. Something with the way homeassistant handles the configuration. I don't know too much about this, maybe we can figure it out.

`Invalid config for 'sensor' from integration 'wnsm' at configuration.yaml, line 28: value should be a string for dictionary value 'device_id', got ['AT0010000000000000001000000000001', 'AT0010000000000000001000000000002', 'AT0010000000000000001000000000003'], please check the docs at https://github.com/DarwinsBuddy/WienerNetzeSmartmeter`

tschoerk, line 6: Elegant solution, but I get the following error when trying to install it via configuration.yaml. Something with the way homeassistant handles the configuration. I don't know too much about this, maybe we can figure it out.

`Invalid config for 'sensor' from integration 'wnsm' at configuration.yaml, line 28: value should be a string for dictionary value 'device_id', got ['AT0010000000000000001000000000001', 'AT0010000000000000001000000000002', 'AT0010000000000000001000000000003'], please check the docs at https://github.com/DarwinsBuddy/WienerNetzeSmartmeter`

tschoerk (changes requested):

DarwinsBuddy:
while going over the config section, I was questioning the approach in general. wouldn't the example provided solve the issue already?

@tschoerk no worries, I've got a backup of your branch as well (additionally to your fork)

tschoerk (approved):
Yeah that way works perfectly and is much simpler than the solutions before. Good one!

PR #262, opened 2024-09-20
Fixed having multiple contracts with multiple zps

DarwinsBuddy:
When switching provider and contract holder a new contract number is given but the zaehlpunkt stays the same. The old one is set to inactive, but is still in the system, if the contracted is not fully deleted. This leads to a situation where the same zaehlpunkt is listed twice under different contracts, once active, once inactive. Currently the updates don't fully check for inactive zaehlpunkte, if the same zaehlpunkt is listed twice. This fix checks all contracts and if the zaehlpunkt is active before adding it, so it gets the correct customer id for retrieving all the data.

I had to revert this PR as I spotted some potential flaws afterwards. @tschoerk maybe you could look over them one more time?

### DIFF
--- a/custom_components/wnsm/api/client.py
+++ b/custom_components/wnsm/api/client.py
@@ -236,17 +236,17 @@ class Smartmeter:
         contracts = self.zaehlpunkte()
         if zaehlpunkt is None:
             customer_id = contracts[0]["geschaeftspartner"]
-            zp = contracts[0]["zaehlpunkte"][0]["zaehlpunktnummer"]
+            zaehlp = contracts[0]["zaehlpunkte"][0]["zaehlpunktnummer"]
             anlagetype = contracts[0]["zaehlpunkte"][0]["anlage"]["typ"]
         else:
             customer_id = zp = anlagetype = None
             for contract in contracts:
-                zp = [z for z in contract["zaehlpunkte"] if z["zaehlpunktnummer"] == zaehlpunkt]
+                zaehlp = [z for z in contract["zaehlpunkte"] if z["zaehlpunktnummer"] == zaehlpunkt and z['isActive']]
                 if len(zp) > 0:
                     anlagetype = zp[0]["anlage"]["typ"]
-                    zp = zp[0]["zaehlpunktnummer"]
+                    zaehlp = zp[0]["zaehlpunktnummer"]
                     customer_id = contract["geschaeftspartner"]
-        return customer_id, zp, const.AnlageType.from_str(anlagetype)
+        return customer_id, zaehlp, const.AnlageType.from_str(anlagetype)
 
     def zaehlpunkte(self):
         """Returns zaehlpunkte for currently logged in user."""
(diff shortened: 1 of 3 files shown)


custom_components/wnsm/api/client.py
```diff
@@ -236,17 +236,17 @@ def get_zaehlpunkt(self, zaehlpunkt: str = None) -> tuple[str, str, str]:
...
         if zaehlpunkt is None:
             customer_id = contracts[0]["geschaeftspartner"]
-            zp = contracts[0]["zaehlpunkte"][0]["zaehlpunktnummer"]
+            zaehlp = contracts[0]["zaehlpunkte"][0]["zaehlpunktnummer"]
             anlagetype = contracts[0]["zaehlpunkte"][0]["anlage"]["typ"]
         else:
             customer_id = zp = anlagetype = None
             for contract in contracts:
-                zp = [z for z in contract["zaehlpunkte"] if z["zaehlpunktnummer"] == zaehlpunkt]
+                zaehlp = [z for z in contract["zaehlpunkte"] if z["zaehlpunktnummer"] == zaehlpunkt and z['isActive']]
```
DarwinsBuddy, line 244: There is a function in `base_sensor` checking for "activeness". Can't we leverage that?

custom_components/wnsm/base_sensor.py
```diff
@@ -93,7 +93,7 @@ def contracts2zaehlpunkte(self, contracts: dict) -> list[dict]:
                 if "zaehlpunkte" in contract:
                     geschaeftspartner = contract["geschaeftspartner"] if "geschaeftspartner" in contract else None
                     zaehlpunkte += [
-                        {**z, "geschaeftspartner": geschaeftspartner} for z in contract["zaehlpunkte"] if z["zaehlpunktnummer"] == self.zaehlpunkt
+                        {**z, "geschaeftspartner": geschaeftspartner} for z in contract["zaehlpunkte"] if z["zaehlpunktnummer"] == self.zaehlpunkt and z['isActive']
```
DarwinsBuddy, line 96: same here. can't we use the `is_active` method?

custom_components/wnsm/config_flow.py
```diff
@@ -37,7 +37,9 @@ async def validate_auth(self, username: str, password: str) -> list[dict]:
         if contracts is not None and isinstance(contracts, list) and len(contracts) > 0:
             for contract in contracts:
                 if "zaehlpunkte" in contract:
-                    zaehlpunkte+=contract["zaehlpunkte"]
+                    for zaehlpunkt in contract["zaehlpunkte"]:              
+                        if zaehlpunkt['isActive']:
```
DarwinsBuddy, line 41: this might prevent adding zaehlpunkte which could be set active later on, and thus would force the user to reimport the whole integration.

DarwinsBuddy (commented):
@tschoerk maybe you can look at it one more time?

tschoerk:
Sorry for late answer had to have a closer look at it. 
The problem was that each zaehlpunkt is only checked once each update and for each check it calls /zaehlpunkte to see if the zaehlpunktnumber is still active. That's why it has to be checked in all loops, in client and base_sensor to make sure the active one is returned. I agree that this is not elegant, but since the stored zaehlpunkt is checked every update, it needs to make sure to return the correct one. I agree though, since it has so many changes for such an edge case, I think it's fine to close.

PR #270, opened 2024-11-08
Fixed get_zaehlpunkt returning empty zaehlpunkt with multiple contracts

### LINKED ISSUE #267: No Data after Update
Hey, 
after yesterdays update to Version 1.5.3 no Data whatsoever is Recieved. No error message, nothing in the logs (at first glance). Message is just: "Unbekannt"
As soon as i roll back to 1.5.2 everything is fine.

Couldn't find the time yet to dig deeper, so no hints as to where the data could get lost. API queries seem to be working.

tschoerk:
When adding or updating a sensor, the zaehlpunkt is checked with the get_zaehlpunkt function. The problem was that the variable for the array, which has the zaehlpunkt details in them, had the same name as the zaehlpunkt itself. So when having multiple contracts/customerIds, the correct zaehlpunkt was overwritten by the zaehlpunkt details, which is now an empty array. All API requests were then done with an empty array and threw errors all around.

Probably fixes #267. The second error [ryanbrown-at](https://github.com/ryanbrown-at) describes, could be due to the recent API changes, which made verbrauch and verbrauch_raw stop working.

### DIFF
--- a/custom_components/wnsm/api/client.py
+++ b/custom_components/wnsm/api/client.py
@@ -241,10 +241,10 @@ class Smartmeter:
         else:
             customer_id = zp = anlagetype = None
             for contract in contracts:
-                zp = [z for z in contract["zaehlpunkte"] if z["zaehlpunktnummer"] == zaehlpunkt]
-                if len(zp) > 0:
-                    anlagetype = zp[0]["anlage"]["typ"]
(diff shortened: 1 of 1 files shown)


tschoerk:
Fixes #267 
The second bug is inherently different and I have opened another issue for it.

DarwinsBuddy (approved):
lgtm

PR #271, opened 2024-11-09
fix API changes

DarwinsBuddy:
* Switch from `/verbrauch` to `/bewegungsdaten` for statistical import
* Add timezone awareness for statistical sensor updates (please @reox can you give it a thorough test and your pair of eyes to confirm, that this solves the issue?)

@tschoerk any feedback is welcome

### DIFF
--- a/custom_components/wnsm/statistics_sensor.py
+++ b/custom_components/wnsm/statistics_sensor.py
@@ -258,25 +258,12 @@ class StatisticsSensor(BaseSensor, SensorEntity):
         _LOGGER.debug("Selecting data up to %s" % now)
         while start < now:
             _LOGGER.debug("Select 24h of Data, using sum=%.3f, start=%s" % (total_usage, start))
-            consumption = await self.get_consumption(smartmeter, start)
-            _LOGGER.debug(consumption)
+            bewegungsdaten = await self.get_bewegungsdaten(smartmeter, start)
+            _LOGGER.debug(bewegungsdaten)
             last_ts = start
             start += timedelta(hours=24)  # Next batch. Setting this here should avoid endless loops
 
-            # Check if this batch of data is valid and contains hourly statistics:
-            if not consumption.get('optIn'):
-                # TODO: actually, we could insert zero-usage data here, to increase the start time
-                # for the next run. Otherwise, the same data is queried over and over.
-                _LOGGER.warning(
-                    f"Data starting at {start} does not contain granular data! Opt-in was not set back then. Skipping.")
-                continue
-
-            # Can actually check, if the whole batch can be skipped.
-            if consumption.get('consumptionMinimum') == 0 and consumption.get('consumptionMaximum') == 0:
-                _LOGGER.debug(f"Batch of data starting at {start} does not contain any consumption, skipping")
-                continue
-
-            if 'values' not in consumption:
+            if 'values' not in bewegungsdaten:
                 # before, this indicated an error. Since 2023-12-15 no more values means that
                 # data is not available (yet) - but could still indicate an error...
                 _LOGGER.debug(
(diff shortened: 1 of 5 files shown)


hwjf:
I'll check your PR with debugging for some time to get some statistic sensor updates of new data as well. However, I suspect the statistics update might still not work as intended.

I assume, since Bewegungsdaten returns data in 15-minute intervals, this line will throw errors 3 times out of 4 for each hour and therefore appends no data, since HomeAssistant needs the timestamp to be from the top of the hour
https://github.com/DarwinsBuddy/WienerNetzeSmartmeter/blob/6189b9a00693e88f843bf9659bc8ae1d42562f00/custom_components/wnsm/statistics_sensor.py#L287-L292

I believe the statistics update needs to be designed similarly to the inital statistical import of data (function _import_bewegungsdaten), where there are two loops, one for summing up the 4 quarterly values to one hourly value and a second loop to append the statistics.
https://github.com/DarwinsBuddy/WienerNetzeSmartmeter/blob/6189b9a00693e88f843bf9659bc8ae1d42562f00/custom_components/wnsm/statistics_sensor.py#L204-L215
https://github.com/DarwinsBuddy/WienerNetzeSmartmeter/blob/6189b9a00693e88f843bf9659bc8ae1d42562f00/custom_components/wnsm/statistics_sensor.py#L228-L231

I'll get back to you with more feedback when the debugging is finished.

tschoerk:
I'll check your PR with debugging for some time to get some statistic sensor updates of new data as well. However, I suspect the statistics update might still not work as intended.

I assume, since Bewegungsdaten returns data in 15-minute intervals, this line will throw errors 3 times out of 4 for each hour and therefore will not append any data (this is done because HomeAssistant needs the timestamp to be from the top of the hour)
https://github.com/DarwinsBuddy/WienerNetzeSmartmeter/blob/6189b9a00693e88f843bf9659bc8ae1d42562f00/custom_components/wnsm/statistics_sensor.py#L287-L292

I believe the statistics update needs to be designed similarly to the inital statistical import of data (function _import_bewegungsdaten), where there are two loops, one for summing up the 4 quarterly values to one hourly value and a second loop to append the statistics.
https://github.com/DarwinsBuddy/WienerNetzeSmartmeter/blob/6189b9a00693e88f843bf9659bc8ae1d42562f00/custom_components/wnsm/statistics_sensor.py#L204-L215
https://github.com/DarwinsBuddy/WienerNetzeSmartmeter/blob/6189b9a00693e88f843bf9659bc8ae1d42562f00/custom_components/wnsm/statistics_sensor.py#L228-L231

I'll get back to you with more feedback when the debugging is finished.

DarwinsBuddy:
@tschoerk I updated it with an idempotent function to spare us some trouble :)

custom_components/wnsm/statistics_sensor.py
```diff
@@ -233,6 +233,26 @@ async def _import_bewegungsdaten(self, smartmeter: Smartmeter):
             _LOGGER.debug(f"Importing statistics from {statistics[0]} to {statistics[-1]}")
         async_import_statistics(self.hass, metadata, statistics)
 
+    def reduce_to_hourly(self, bewegungsdaten):
+        hourly_data = defaultdict(float)
+        for entry in bewegungsdaten['values']:
+            ts = dt_util.parse_datetime(entry['zeitpunktVon'])
+            hour_start = ts.replace(minute=0, second=0, microsecond=0)
+            hourly_data[hour_start] += entry['wert']
+
```
reox, line 242: Maybe add a check here if always four quarters were aggregated? Not that we can do much if that happens...

DarwinsBuddy: I think this would overcomplicate the code. sure if the API is not reporting the right amount of data points, we would simply store it as we get it. but as you said, nothing which we can do anyhow here. :shrug:

DarwinsBuddy:
@reox I fixed the type mismatch.
I don't know if this spike is not from some faulty import beforehand.

Can you check by removing all of your data (i.e. resetting) and re-import?
I can not reproduce this behaviour :/

tschoerk (approved):
Tested it over 24 hours with fresh data. Everything works from my side. Couldn't recreate the massive influx of data at midnight @reox and others have reported though.

reox:
Seems like this was caused by partial import of the data, before I fixed the errors.
The data of yesterday was imported correctly. I'll (if I don't forget) also start a full import today. But I think the PR is good!

DarwinsBuddy:
After playing around with the import code I extracted some fixtures for mocking the bewegungsdaten from my jupyter notebook and integrated into our tests to easily mock the data from `/bewegungsdaten`

I'm merging this now and releasing just to get this out. Let's see if more people have these issues with the "peak import"

PR #274, opened 2024-11-14
Fix/improve multiple Zaehlwerke/OBIS Codes

### LINKED ISSUE #272: historical_data throws error with multiple Zaehlwerke/OBIS Codes
[log.txt](https://github.com/user-attachments/files/17709291/log.txt)

_Originally posted by @ryanbrown-at in https://github.com/DarwinsBuddy/WienerNetzeSmartmeter/issues/267#issuecomment-2469328272_

The live_sensor, which saves the current meter readings, uses a call to historical_data to get the current meter readings. This function works only with one zaehlwerk, since it has usually only one and we didn't know yet, how to data with multiple zaehlwerke looks like.

Opening a new issue from #267 since ryanbrown-at hat multiple different errors.

tschoerk:
Implemented the ability to check different valid OBIS codes. Wiener Netze is currently using two OBIS codes as default for meter readings 1-1:1.8.0 for consumption and 1-1:2.8.0 for production. Apparently they started using 1-1:1.9.0 especially for people using heat pumps, which is the sum of 1-1:1.8.1 and 1-1:1.8.2. Since there was basically no checking of valid OBIS codes, except if there's only one present (the amount was 3 in the case of heat pumps, therefore an error was thrown), this heavily improves upon this.

This fixes #272

### DIFF
--- a/custom_components/wnsm/api/client.py
+++ b/custom_components/wnsm/api/client.py
@@ -405,14 +405,28 @@ class Smartmeter:
         """
         Find and validate data with valid OBIS codes from a list of zaehlwerke.
         """
+        
+        # Check if any OBIS codes exist
+        all_obis_codes = [zaehlwerk.get("obisCode") for zaehlwerk in zaehlwerke]
+        if not any(all_obis_codes):
+            raise SmartmeterQueryError("No OBIS codes found in the provided data.")
+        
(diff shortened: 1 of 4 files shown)


DarwinsBuddy (approved):
lgtm

PR #275, opened 2024-11-14
fix(statistics): Remove reduce_to_hourly and warning

DarwinsBuddy:
I decided to remove the code for reducing everything to one hour splices alongside a warning for importing qh datapoints.

Merged the import support into one generic method.

 @reox @tschoerk please let me know what you think.

### DIFF
--- a/custom_components/wnsm/statistics_sensor.py
+++ b/custom_components/wnsm/statistics_sensor.py
@@ -119,13 +119,13 @@ class StatisticsSensor(BaseSensor, SensorEntity):
             if not self.is_last_inserted_stat_valid(last_inserted_stat):
                 # No previous data - start from scratch
                 _LOGGER.warning("Starting import of historical data. This might take some time.")
-                await self._import_bewegungsdaten(smartmeter)
+                await self._initial_import_statistics(smartmeter)
             else:
                 start_off_point = self.prepare_start_off_point(last_inserted_stat)
                 if start_off_point is None:
                     return
                 start, _sum = start_off_point
-                await self._import_statistics(smartmeter, start, _sum)
+                await self._incremental_import_statistics(smartmeter, start, _sum)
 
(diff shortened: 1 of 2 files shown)


custom_components/wnsm/statistics_sensor.py
```diff
@@ -144,200 +144,78 @@ async def async_update(self):
...
+                # Usually this means that the measurement is not yet in the WSTW database.
+                continue
+            reading = Decimal(value['wert'] * factor)
             if ts.minute % 15 != 0 or ts.second != 0 or ts.microsecond != 0:
                 _LOGGER.warning(f"Unexpected time detected in historic data: {value}")
-            if (ts_to - ts) != timedelta(minutes=15):
-                _LOGGER.warning(f"Unexpected time step detected in historic data: {value}")
             dates[ts.replace(minute=0)] += reading
+            if value['geschaetzt']:
+                _LOGGER.debug(f"Not seen that before: Estimated Value found for {ts}: {reading}")
```
tschoerk, line 211: Just for info: geschaetzt happens if the smart meter is off the grid for some time, so for example a power outage. I had this one happen couple weeks ago, where the power went out for 3 hours in our area, so for those 3 hours the usage was of course 0 but since they couldn't know for sure it had the attribute geschaetzt set to true. I don't know if this helpful in any way, but the more you know :)

tschoerk (approved):
Tested it with completly new setup and works fine. Removed the last couple days and it worked fine as well. So everything good from my side!

custom_components/wnsm/statistics_sensor.py
DarwinsBuddy: Good to know. Thanks for this insight. :)  I think that we will just have to blindly import this data, as it's a) lost also from their side, b) we cannot expect that this data is recovered any time soon, and c) even if, we cannot retrospectively fill gaps as of the limitations of home assistant's recorder.

PR #277, opened 2024-11-17
refactor(sensors): Merge live + statistics sensor into WNSMSensor

DarwinsBuddy:
Hi @tschoerk and @reox 

I invested some time to refactor the sensors into one: `WNSMSensor`
It's basically the live sensor, but it's calling additionally an `importer`, which bears the logic from `StatisticsSensor`.
I used following integration (included in the core integrations) as a blueprint: [Elvia](https://github.com/home-assistant/core/tree/dev/homeassistant/components/elvia)

The statistics sensor I sunset and marked deprecated it. After releasing this I'd properly document this in the docs and the release notes and would suggest deleting the statistics sensor i.g. A later release (some time later) would then remove the `StatisticsSensor` from the code base eventually.

The historical data is exposed to home assistant differently.
Instead of
`sensor.atXXXXXXXXXXXXXXX_statistics`
it will now be offered under
`wnsm:atXXXXXXXXXXXXXXX`
which is valid according to following line `https://github.com/home-assistant/core/blob/6f947d2612374cb5870729ddec864e2fd59200ef/homeassistant/components/recorder/statistics.py#L326`

Following things are to be considered:

* Deprecate statistics sensor and disable on next update Historical data is now reported to wnsm:atXXXXXXXX
* Breaking changes: energy dashboard has to be reconfigured with the new historical data stream

I'd really appreciate your input on this.
Be careful, you might have to do a data wipe if you want to switch between this branch and `main`


tschoerk:
Wow very nice work! :)
Let me get on to testing this for a couple days, with all the different data quirks we know can happen and get back to you!

reox:
Nice! Unfortunately, I'm very short on time right now and probably cannot do a full review in the next weeks :(
Just one question: Does it now handle states + statistics in a sane way, i.e., is this now supported by HA? Back then, when we decided to have two sensors, we had the problem that having a state would result in broken statistics.

DarwinsBuddy:
@reox yes, I guess so. Since we do not necessarily need an explicit sensor to save something with recorder. We just have to use the domain of our integration (`wnsm`) and update the stats there on update of our sensor.

This means that we have 2 histories produced.
* One by the `WNSMSensor` (which is a) showing the last meter reading from the API and b) producing for each sucessfull update a datapoint in its respective recorder stream under `sensor.atXXXXXXXX`)
* One by the import mechanism which produces another recorder stream under `wnsm:atXXXXXXXXXX`


So only one sensor, but two histories, hence no need for a separate sensor that shows no status whatsoever and would conflict with the dates imported (since they wrote into the same recorder stream)

DarwinsBuddy:
@reox @tschoerk Any update on your findings? I'd like to release a new version containing a fix for #279 and considering adding all of this to it (iff there is no second thoughts from your side)

reox:
@DarwinsBuddy sorry, still hadn't had the time to look at the code properly :( But I guess it's fine? Not having the two entities sounds like a good idea!

tschoerk:
Sorry for getting back so late. The API seemed to have a lot of problems in the last week, since I got a ton of error messages regarding server problems (Could not obtain API, No response, etc.). 

What I did notice though is that the update of data takes a lot longer than the old method. I get the home assistant warning `Update of sensor is taking over 10 seconds` every hour for each sensor. I looked through the code and noticed two things which probably could be improved upon:
1. For both updates the whole login process is done twice. So it logins for the meter readings and afterwards another login for the statistics update, although they happen basically in the same process.
2. Both meter readings and statistics are checked now every hour, even if the newest data from last day was already received. For meter readings there was never a method implemented to prevent that but the old statistics update always waited 24 hours before starting with the calls (though this led to the import bug with the big spikes).

Both could save a ton of API calls which would speed up the process and lead to fewer errors overall (maybe some server errors I got were due to rate limiting, but I can't test that).


Regarding the NACHTSTROM fix, could idea with the error message just showing the label, so we can fix it faster in the future.

reox:
> For both updates the whole login process is done twice. So it logins for the meter readings and afterwards another login for the statistics update, although they happen basically in the same process.

I remember that the login gives a token and a refresh token. I believe it should be possible to store both and call the refresh every now and then, to not run the full login process again. I think last time we discussed that, the problem was that there was no good way to store such information in home assistant?

tschoerk:
> > For both updates the whole login process is done twice. So it logins for the meter readings and afterwards another login for the statistics update, although they happen basically in the same process.
> 
> I remember that the login gives a token and a refresh token. I believe it should be possible to store both and call the refresh every now and then, to not run the full login process again. I think last time we discussed that, the problem was that there was no good way to store such information in home assistant?

Yeah I read that discussion and I think they are only valid for 30 minutes and would need a seperate process to refresh, which is not possible in HA (at least at the time of your discussion).

In this case I would just use the tokens which were already provided from the login in the meter reading update and pass the AsyncSmartmeter object to the importer. I will comment on the lines, so you know what I mean.

custom_components/wnsm/wnsm_sensor.py
```diff
@@ -0,0 +1,108 @@
...
+            zaehlpunkt_response = await async_smartmeter.get_zaehlpunkt(self.zaehlpunkt)
+            self._attr_extra_state_attributes = zaehlpunkt_response
+
+            if async_smartmeter.is_active(zaehlpunkt_response):
+                # Since the update is not exactly at midnight, both yesterday and the day before are tried to make sure a meter reading is returned
+                reading_dates = [before(today(), 1), before(today(), 2)]
+                for reading_date in reading_dates:
+                    meter_reading = await async_smartmeter.get_meter_reading_from_historic_data(self.zaehlpunkt, reading_date, datetime.now())
+                    self._attr_native_value = meter_reading
+                importer = Importer(self.hass, self.username, self.password, self.zaehlpunkt, self.unit_of_measurement, self.granularity())
```
tschoerk, line 97: Here the AsyncSmartmeter object which was initialized on Line 86 and did the login on line 87 could be passed to the importer, to save the second login.

DarwinsBuddy:
Thank you so much for your feedback,
That's very insightful. I missed these points as I refactored.
As soon as I have time I'll try to implement the mentioned improvements.

DarwinsBuddy:
@tschoerk I added a check if the session already logged in and is still valid (according to the expiration)
The AsyncSmartmeter is now past to the importer, using the same asyncio lock to prevent race conditions.
What do you think?

DarwinsBuddy:
I'd merge this PR now, since we have a bunch of integrations failing to logging in (due to the rate limits + captcha on logins from WienerNetze).

Let's hope for the best. :)

tschoerk:
> I'd merge this PR now, since we have a bunch of integrations failing to logging in (due to the rate limits + captcha on logins from WienerNetze).
> 
> Let's hope for the best. :)

Sorry didn't have too much time at my hand, but looks good and I think everyone with 1 zaehlpunkt won't have login troubles any longer.
It seems though that WienerNetze instantly throws the captcha after the first login. So everyone with multiple zaehlpunkte gets the login error.

Some ideas/enhancement going forward regarding the rate limiting:
1. Add specific error if it really is a captcha page, so we can know for sure that this is the problem.
2. Check the last update time of the sensor and wait at least 24 hours (or something like 5 AM, whatever is earlier) before logging in again, so there's no need to even login after new data has arrived. Wiener Netzte provides the data in 24 hour batches anyway, so once we have it for the last day we can wait till the next morning hours and only then check hourly if nothing is avalaible.
3. I don't know if this is even possible, but maybe we have to switch to a synchronous update of the sensors and use 1 login for every zaehlpunkt available. This should completly eliminate all captcha issues.

PR #309, opened 2025-05-24
fix-login-add-pkce

cretl:
Add PKCE (Proof Key for Code Exchange) for the login to fix the login error

### DIFF
--- a/custom_components/wnsm/api/client.py
+++ b/custom_components/wnsm/api/client.py
@@ -9,6 +9,10 @@ import requests
 from dateutil.relativedelta import relativedelta
 from lxml import html
 
+import base64
+import hashlib
+import os
+
 from . import constants as const
 from .errors import (
     SmartmeterConnectionError,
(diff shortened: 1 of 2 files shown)


joker1319:
tested and works fine

cretl:
I made a methodical mistake that caused this problem with more than one meter. I'll try to fix it.

gluckb:
It worked, thank you.

Rgasd:
Can someone merge this fix...hopefully or is this repo not maintained anymore?

spinnaker1:
Can someone implement this fix into other repository? https://github.com/platysma/vienna-smartmeter/issues/181

tschoerk (commented):
Thanks a lot for the PR and quick fix. Looks good and seems to be working. Please also update tests covering the new PKCE logic then we can merge.

cretl:
I couldn't find an easy way to fix the build check. The only solution I found was to pass a fixed PKCE code verifier to the Smartmeter class. This is only necessary for the GitHub check and not for the script itself. I'll try to find a better solution.

tschoerk:
> I couldn't find an easy way to fix the build check. The only solution I found was to pass a fixed PKCE code verifier to the Smartmeter class. This is only necessary for the GitHub check and not for the script itself. I'll try to find a better solution.

You're right—this is tricky. The test fails because mock_token uses post_data_matcher, which compares the full request body, and the code_verifier is randomly generated at runtime, so even if adding a fixed code_verifier to the mock requests, it'd fail since the actual request is always different.

One idea to fix this is to modify mock_token (or post_data_matcher but this is used for many different request checks, so I don't think it's the cleanest solution to do it there) to check only for the presence of code_verifier in both the actual and expected request bodies, rather than comparing the exact value. This would preserve the random nature of the PKCE flow while still ensuring the structure of the request is correct.

Sinuati:
it works! thank you!

cretl:
> You're right—this is tricky. The test fails because mock_token uses post_data_matcher, which compares the full request body, and the code_verifier is randomly generated at runtime, so even if adding a fixed code_verifier to the mock requests, it'd fail since the actual request is always different.

The tests also fail because the URLs created by the mock_login_page and load_login_page functions don't match. The latter function creates a different code verifier. Due to this the code_challenge parameter in the URLs is differs.
As a "dirty fix," I added an input_code_verifier variable to client.py to be able to use the same code_verifier.

The tests should now pass.

I would suggest merging this PR to fix the main issue with the login and add a "real" fix for the tests later.

Nihvel:
Forcefully copied the updated version over, and the integration is working again! Good job! +1

PR #326, opened 2026-01-05
Handle missing unit of measurement in API response

cretl:
When there is no new data to import, the API responds with the value "null" for the key "descriptor.einheit"
API response: 
``` json 
{
  "descriptor": {
    "geschaeftspartnernummer": "***",
    "zaehlpunktnummer": "AT***",
    "rolle": "V001",
    "aggregat": "NONE",
    "granularitaet": "D",
    "einheit": null
  },
  "values": []
}
```
When the value for the key "descriptor.einheit" is "null", there are normally no values in the response. 

This PR changes the following:

1. [importer.py]: First, check the "values" of the API response, then the value of bewegungsdaten["unitOfMeasurement"].
2. [AsyncSmartmeter.py]: Set the key for descriptor.einheit to an empty string if the API responds with the value "null."
3. [importer.py]: Handle the situation when the value of bewegungsdaten['unitOfMeasurement'] (equivalent to the JSON key "descriptor.einheit") is empty.

This PR should resolve issue #324.

### DIFF
--- a/custom_components/wnsm/importer.py
+++ b/custom_components/wnsm/importer.py
@@ -160,16 +160,19 @@ class Importer:
         if 'unitOfMeasurement' not in bewegungsdaten:
             _LOGGER.debug("unitOfMeasurement does not exist. Aborting import!")
             return
-        elif bewegungsdaten['unitOfMeasurement'] == 'WH':
+
+        if bewegungsdaten['unitOfMeasurement'] == '':
+            _LOGGER.debug("WienerNetze API response does not contain a unit of measurement. Aborting import!")
+            return
+
+        if 'values' not in bewegungsdaten or bewegungsdaten['values'] is None or len(bewegungsdaten['values']) == 0:
+            _LOGGER.warning(f"Malformed bewegungsdaten {bewegungsdaten}. Aborting import!")
+            return
+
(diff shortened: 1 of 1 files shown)


reox:
lgtm!

leobowenwang:
When will this be merged, is this project still actively maintained?

custom_components/wnsm/AsyncSmartmeter.py
```diff
@@ -150,6 +150,9 @@ async def get_bewegungsdaten(self, zaehlpunkt: str, start: datetime = None, end:
         if "Exception" in response:
             raise RuntimeError(f"Cannot access bewegungsdaten: {response}")
         _LOGGER.debug(f"Raw bewegungsdaten: {response}")
+        if response.get('descriptor', {}).get('einheit') is None:
```
DarwinsBuddy: let's remove that (see comment in importer)

DarwinsBuddy:
I created another PR with an alternative solution.
https://github.com/DarwinsBuddy/WienerNetzeSmartmeter/pull/337
wanna catch up here https://github.com/DarwinsBuddy/WienerNetzeSmartmeter/issues/324 ?

PR #331, opened 2026-02-15
Major Update

klabir:
The main change is the addition of daily values, which are often available earlier than the total consumption. New sensors and statistics have been introduced to assign values to the correct reading date. Additional attributes have been added to the sensors. The authentication flow has been redesigned to reduce the risk of blocking. The YAML configuration has been removed and replaced with a UI-based configuration for the scan interval. Backward compatibility should be maintained, since the main sensors for each metering point remain unchanged.

I hope you will consider taking this over so that there are no redundant HACS custom components. Please feel free to reach out if you have any questions.

Done by Codex 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,9 +64,9 @@ For the **Energy Dashboard (grid consumption)**, prefer `wnsm:<slugified-zaehlpu
 Use one series per card (do not mix DAY and snapshot in the same statistics-graph card).
 
 - **Daily values (DAY):**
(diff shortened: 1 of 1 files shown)


klabir:
one bug needs to be fixed

W-M-B:
Hi @klabir , I was just waiting for someone to use LLM to provide updates. here we go.
from what I see you committed quite a bit of changes that would require quite some testing.
Can you share what has been tested with unit tests etc. 
I would suggest to use this as a test version and get some testing done by a wider user group. just my cents.

PR #332, opened 2026-02-15
Add daily values + reading-date fixes + UI scan interval + auth flow

klabir:
Hello 

The main change is the addition of daily values, which are often available earlier than the total consumption. New sensors and statistics have been introduced to assign values to the correct reading date. Additional attributes have been added to the sensors. The authentication flow has been redesigned to reduce the risk of blocking. The YAML configuration has been removed and replaced with a UI-based configuration for the scan interval. Backward compatibility should be maintained, since the main sensors for each metering point remain unchanged.

I hope you will consider taking this over so that there are no redundant HACS custom components. Please feel free to reach out if you have any questions.

Done by Codex


klabir:
will fix other things as well

raoulbhatia:
@klabir this is getting boring 🫠

PR #338, opened 2026-03-17
Add 'WANDLER' to AnlagenType from_str cases

upchui:


DarwinsBuddy:
@upchui I suspect that's a type definitely existing and you know what you're doing. :crossed_fingers: I will merge, since I can't fact check it unfortunately.

upchui:
his is what it looks like when you have a "Wandlerzähler" with Wiener Netze.

image

Here is a short article:

https://www.inexogy.com/blog/was-sind-wandlerzaehler-und-wandlermessung/


Thanks for merging it. It’s quite possible that I’m the first and only one so far who has a “Wandlerzähler” and is using this Home Assistant integration for it.

PR #343, opened 2026-04-27
added Energiegemeinschaft

AndiVienna91:
added sensors and statistics for roles G001 and G003


raoulbhatia:
@AndiVienna91 having that many changes in your PR doesn't look right.
What do you think?

reox:
Why did you basically remove all comments and docstrings? 🤔

DarwinsBuddy:
my question exactly.
If you want to contribute to this project, please check your output after vibe coding.

This doesn't seem like a minimal change for the issue addressed.

Your intention is very much welcome, since it's a feature that has been requested oftentimes, but please try to make this most minimal so we can actually check it.

Removing comments without reason is not a good way of contributing.

DarwinsBuddy (changes requested):
take care of formatting and don't remove comments

make this a minimal change please

AndiVienna91:
I had the code cleaned up by an AI after some experimentation, the comments apparently got removed in the process.

They have now been restored.

DarwinsBuddy (changes requested):
I really value your input, but this is still not minimal.
I do get that using LLMs for this kind of work is appealing and useful, but if everyone changes comments arbitrarily there is no such sense in having comments.

A PR claming of adding something should not change something that is unrelated.
If that's not possible to prompt to an AI successfully, I suggest going over the changes one by one in your favourite Difftool and prepare this PR properly by squashing.

AndiVienna91:
No offense, but I’m not going to keep reworking a piece of functioning code just to match your preferences.

After trying several different approaches, this version turned out to be the most reliable and stable one from my perspective, and it meets my requirements.

You’re welcome to extract and integrate only what you consider essential into your own code. If it works, the community benefits—if not, then so be it.

That said, I do wonder why these features haven’t already been implemented if there’s clearly demand for them.

In any case, I’m stepping out at this point. Wishing you all the best moving forward.

reox:
I can understand your frustration, but on the other side is also frustrating for the reviewers.
The thing is: using LLMs shifts the burden drastically from implementation to review. And not implementing minimal changes increases the review time and if things get changed that do not remotely have something to do with the intent of the commit, one has to ask why...

> No offense, but I’m not going to keep reworking a piece of functioning code just to match your preferences.

well, removing comments or completely rewriting them is not good. I mean okay, you added now docstrings for previously undocumented functions - credits for that - but you removed, for example, pylint statements, which have a specific meaning. These are not just preferences but information for the test-system.
And "functioning code" != "good code".

> After trying several different approaches, this version turned out to be the most reliable and stable one from my perspective, and it meets my requirements.

yes, it meets _your_ requirements. But remember, there are a lot more people using this tool and when programming on such projects, you have to keep their requirements in mind too.

> That said, I do wonder why these features haven’t already been implemented if there’s clearly demand for them.

Several reasons, for example:
* time - this is after all, a project that is developed in the free time
* no possibility to test - not everyone has an energiegemeinschaft and thus implementing something blind is not easy
* no official API docs - everything needs to be reverse engineered, adding more time on top

DarwinsBuddy:
@AndiVienna91 Many thanks again to your initiative and the provided contribution.

Currently Open Source is undergoing a challenge with vibe coded code and contributions that are hard to maintain and verify.

Although we are a relatively small project compared to others that struggle significantly more (e.g. curl), we do still need to ensure that what someone is contributing is
a) supporting the cause (in your case introducing a new feature, namely Energiegemeinschaft)
b) not breaking the current functionality, i.e. regression

Usually when a change has 1000+ loc changed this is already a bad sign for such a change, as it should potentially involve way less changes for such an addition, unless we have to majorly refactor something in order to enable this feature.

Unfortunately, I don't see this being the case here.

In fact, your contribution deletes potentially important but still unrelated, to your change, comments and also throws away lint information.

As @reox pointed out, there is a small group of people that maintain this project pro bono, with no compensation whatsoever. If you want to contribute something valuable, please be considerate of the time of the maintainers, since they are the gatekeepers of breaking changes.

The change that you provided might "work on your machine", but it's a) not easy to verify for us (since we do not have access to accounts from Energiegemeinschaft) and b) it's extremely hard to read and c) cannot be merged as is, as it is introducing more problems due to missing lint hints and missing comments that serve all a purpose.

If you do not understand the importance of these topics, I'd suggest that you educate yourself about it. I'm happy to share some valuable material on specific topics if needed.

But then also: only because you can instruct an LLM to vibe code something, doesn't necessarily mean that you can provide it with the necessary quality standards that are a non-negotiable to the open source community.

If there is someone else that wants to rejuvenate this contribution and carry it over the finish line to make this a minimal change to introduce, we all would be happy for it.

Thank you for your interest and your effort. I hope that doesn't demotivate you to keep on learning.

W-M-B:
I also have an EC as consumer and happy to join in. Timing is a little difficult right now.

Issue #349, opened 2026-05-26
Every 1-3 weeks no data


Swimmer71:
Hi,
it happens every some weeks, that i get no data or not all data. I found out, when I re-download this WNSM in HACS, and after reboot, the data is back again. So, what could be the problem?
At what time this integration fetches data and how often? Is there a way (action or so) to trigger the data-fetching process?

mariomurrent-softwaresolutions:
Same here - data loading is stuck after a week

gerihouse:
same here

W-M-B:
Do you have logs to share? Any special setup?

In the Webportal you can see everything, right?

Swimmer71:
right now it works, have to wait until next problem. Were can I find the log you need?

My config:
I have 3 smartmeter with 3 separate account (mine, my son, my mother) and I all integrate them in HA. Until 2 weeks ago I integrate them via GUI. Then I found the possibility to do this via configuration.yaml in section sensor:
```
  - platform: wnsm
    username: !secret wnsm_c_u
    password: !secret wnsm_c_p
    device_id: !secret wnsm_c_zp

  - platform: wnsm
    username: !secret wnsm_a_u
    password: !secret wnsm_a_p
    device_id: !secret wnsm_a_zp

  - platform: wnsm
    username: !secret wnsm_g_u
    password: !secret wnsm_g_p
    device_id: !secret wnsm_g_zp

```
Now it does not appear in the "integrations" but this is maybe normal as it was setup via yaml. Now I test, if the problem does not appear with this method.

But, at what time the integration picks the data from WN? And is there an action to do this manually?

r14g:
same...
Logs bitte sehr:

```
Logger: homeassistant.components.sensor
Source: helpers/entity_platform.py:848
Integration: Sensor (documentation, issues)
First occurred: 12:19:06 (2 occurrences)
Last logged: 12:19:07

wnsm: Error on device update!
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 848, in _async_add_entity
    await entity.async_device_update(warning=False)
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1373, in async_device_update
    await self.async_update()
  File "/config/custom_components/wnsm/wnsm_sensor.py", line 99, in async_update
    await importer.async_import()
  File "/config/custom_components/wnsm/importer.py", line 106, in async_import
[...]

Issue #287, opened 2024-12-05
No hourly data


ralpho1337:
Hi guys,

thanks for creating this integration.

I unfortuantely do not get any hourly data. The daily meter reading works fine.

Activating the debugging while initiating the integration reveals the following (i removed my geschaeftspartner and zaehlpunktnummer). Any idea why the bewegungsdaten API does not have any data for me?

Thanks
Ralph

API Request: https://service.wienernetze.at/sm/api/user/messwerte/bewegungsdaten?geschaeftspartner=xxx&zaehlpunktnummer=AT001000000000000000xxxxx&rolle=V002&zeitpunktVon=2024-06-14T22%3A00%3A00.000Z&zeitpunktBis=2024-12-05T23%3A59%3A59.999Z&aggregat=NONE

API Response: {
  "descriptor": {
    "geschaeftspartnernummer": "xxxx",
    "zaehlpunktnummer": "AT001000000000000000xxxx",
    "rolle": "V002",
    "aggregat": "NONE",
    "granularitaet": "QH",
    "einheit": "KWH"
  },
  "values": []
}
2024-12-05 13:18:24.783 DEBUG (MainThread) [custom_components.wnsm.AsyncSmartmeter] Raw bewegungsdaten: {'descriptor': {'geschaeftspartnernummer': 'xxxx', 'zaehlpunktnummer': 'xxxxxx', 'rolle': 'V002', 'aggregat': 'NONE', 'granularitaet': 'QH', 'einheit': 'KWH'}, 'values': []}
2024-12-05 13:18:24.783 DEBUG (MainThread) [custom_components.wnsm.importer] Mapped historical data: {'customerId': 'xxxx', 'zaehlpunkt': 'xxxxx', 'role': 'V002', 'aggregator': 'NONE', 'granularity': 'QH', 'unitOfMeasurement': 'KWH', 'values': []}
2024-12-05 13:18:24.783 DEBUG (MainThread) [custom_components.wnsm.importer] Batch of data starting at 2024-06-14 22:00:00+00:00 does not contain any bewegungsdaten. Seems there is nothing to import, yet.

tschoerk:
Do you have hourly data at your smartmeter web view [here](https://smartmeter-web.wienernetze.at/verbrauchsdaten/zaehlerdaten)?
Have you opted in to quarter hour updates at your smartmeter or only hourly?

Issue #215, opened 2024-01-26
Error on device update


antonmosich:
I'm sorry, I'm really not deep enough into home-assistant to provide a proper bug report, but for me it only shows "Unknown" for the added entity, and isn't able to load any data from the API. Looking at my logs I seem to have 3 errors with this plugin.
The first has this Traceback:

```
Traceback (most recent call last):
  File "/nix/store/gx2n17v083h4p8jf34cibp0qz4mrwxph-homeassistant-2024.1.3/lib/python3.11/site-packages/homeassistant/helpers/entity_platform.py", line 573, in _async_add_entity
    await entity.async_device_update(warning=False)
  File "/nix/store/gx2n17v083h4p8jf34cibp0qz4mrwxph-homeassistant-2024.1.3/lib/python3.11/site-packages/homeassistant/helpers/entity.py", line 1216, in async_device_update
    await self.async_update()
  File "/var/lib/hass/custom_components/wnsm/statistics_sensor.py", line 123, in async_update
    await self._import_bewegungsdaten(smartmeter)
  File "/var/lib/hass/custom_components/wnsm/statistics_sensor.py", line 187, in _import_bewegungsdaten
    recording = await self.get_bewegungsdaten(smartmeter)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lib/hass/custom_components/wnsm/base_sensor.py", line 156, in get_bewegungsdaten
    response = await self.hass.async_add_executor_job(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/asiphbpiy2gmidfm3xbwcikayhs66289-python3-3.11.7/lib/python3.11/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lib/hass/custom_components/wnsm/api/client.py", line 506, in bewegungsdaten
    if data["descriptor"]["zaehlpunktnummer"] != zaehlpunkt:
       ~~~~^^^^^^^^^^^^^^
KeyError: 'descriptor'
```
The second one just says unable to load consumption without further information.
The third one has the following information (anonymized) 
```
[...]

PR #355, opened 2026-06-20
Gracefully handle bewegungsdaten responses without a "descriptor"

microcd-dev:
## Problem

When WienerNetze returns a `bewegungsdaten` response **without** a `descriptor`
key (an error/empty body, or simply no movement data for the requested period
yet), `Smartmeter.bewegungsdaten()` raises a raw `KeyError: 'descriptor'`:

```
  File ".../custom_components/wnsm/api/client.py", line 636, in bewegungsdaten
    if data["descriptor"]["zaehlpunktnummer"] != zaehlpunkt:
       ~~~~^^^^^^^^^^^^^^
KeyError: 'descriptor'
```

Neither `Importer.async_import()` nor `WNSMSensor.async_update()` catch
`KeyError`, so this surfaces as `wnsm: Error on device update!` and the sensor
goes **unavailable until a manual restart / reinstall**. This is the crash
reported in #349 (and earlier in #287 / #215).

## Root cause

`bewegungsdaten()` indexed `data["descriptor"]["zaehlpunktnummer"]`
unconditionally. The reverse-engineered WN endpoint does not always include a
`descriptor` (it can return an error body, or an empty payload when no data is
available yet).

## Fix

- **`api/client.py`** — guard the response and raise `SmartmeterQueryError`
  (consistent with the existing `historical_data()` check) when `descriptor` is
  absent, instead of a raw `KeyError`.
- **`importer.py`** — catch `SmartmeterError` in `async_import()` so a transient
  WN data gap skips the statistics import for that cycle (logged as a warning)
  and the sensor recovers automatically on the next update, rather than failing
  the whole device update and going unavailable.

Net effect: a missing/empty `bewegungsdaten` response no longer takes the sensor
down; it is treated as "no data this cycle" and recovers on its own.

## Tests

- Added `test_bewegungsdaten_no_descriptor` plus a `no_descriptor` option to the
  `bewegungsdaten` test fixture, asserting a `SmartmeterQueryError` is raised
  instead of a `KeyError`.
- Full suite green locally (`coverage run -m pytest`), flake8 syntax/undefined
  gate clean.

Relates to #349.


W-M-B:
@reox  @cretl @tschoerk who can review the pr?
I have installed it now and provide feedback how it worked for my installation.

W-M-B:
replaced the 3 changed files restarted homeassistant and all missing data is up2date.
dont know about side effects yet, but for me the commit works.

DarwinsBuddy (approved):
lgtm

PR #362, opened 2026-07-09
Fall back to daily granularity when the quarter-hour feed is empty (#361)

### LINKED ISSUE #361: Unit of measurement is None
Mit neuester Version

Dieser Fehler stammt von einer benutzerdefinierten Integration

Logger: custom_components.wnsm.importer
Quelle: custom_components/wnsm/importer.py:169
Integration: WienerNetzeSmartmeter (Dokumentation, Probleme)
Erstmals aufgetreten: 03:39:28 (2 Vorkommnisse)
Zuletzt protokolliert: 09:39:27

Unit of measurement is None! Aborting import...

Sgoettschkes:
## Problem

Fixes #361.

When a meter is opted into 15-minute values (Viertelstundenwerte), its role flips from `V001` (daily) to `V002` (quarter-hour), and `WNSMSensor.granularity()` follows the account into `QUARTER_HOUR`. WienerNetze can return an empty response for the V002 feed (`einheit: null`, `values: []`) even while the daily `V001` feed is still populated. The importer then hits:

```
Unit of measurement is None! Aborting import...
```

and writes nothing, so the energy dashboard stops updating.

## Change

`Importer._import_statistics` now fetches through a small helper `_get_bewegungsdaten()` that:

1. requests the configured granularity (unchanged default behaviour), and
2. if the quarter-hour response has no unit of measurement, retries once at daily (`ValueType.DAY` / V001).

So 15-minute data is used whenever WienerNetze serves it, and daily is used as a fallback otherwise. The existing `unitOfMeasurement is None` guard remains as a final backstop — if even the daily feed is empty, it aborts cleanly as before. Downstream aggregation is unchanged, since values are bucketed hourly regardless of source granularity.

The fallback only triggers when the requested granularity isn't already daily **and** the response has no unit, so a normal incremental cycle with no newly-finalised readings doesn't cause an extra API call.

## Testing

Verified on a live Home Assistant instance whose meter was opted into 15-minute values and hitting #361: after the change the integration imports again — using quarter-hour data when WienerNetze serves it, and falling back to daily when the V002 feed comes back empty.


Norus6:
This seems like a good fix. I have that issue too on 1.8.1. No more data coming in

DarwinsBuddy:
Thank you for your contribution, unfortunately after testing with my data, neither @kbrozek nor your PRs were rock solid.

I implemented a not completely accurate, but hopefully more robust approach for the historical import.

Thank you very much for your contribution still, as it pointed me towards the issue and opened up the understanding for its complexity. :heart:

PR #365, opened 2026-08-25
Prefer the active contract when resolving a zaehlpunkt

okonet:
## Problem

`Smartmeter.get_zaehlpunkt()` loops over every contract without breaking, so when the same zaehlpunkt is listed under more than one `geschaeftspartner` it returns the **last** match instead of the active one.

This happens after a move or a change of supplier: the meter stays the same, but it shows up under both the expired and the current contract. When the expired contract is returned last, every `bewegungsdaten` query goes out with its `geschaeftspartner`.

The API does not fail in that case. It answers `200` with a `descriptor` whose `einheit` is `None` and an empty `values` list, so `Importer` bails out at:

```
Unit of measurement is None! Aborting import...
```

No statistics are ever written, the sensor sits at `unknown`, and nothing else in the log points at the cause. It looks like an auth or account problem rather than the wrong customer id being used.

## Fix

Break on the first contract whose zaehlpunkt is `isActive`, and fall back to the previous last-match behaviour when none is active — so single-contract setups are unaffected.

## Verification

Checked against the live API for a zaehlpunkt listed under two contracts:

| geschaeftspartner | `einheit` | quarter-hour values |
| --- | --- | --- |
| active | `KWH` | 16796 |
| expired | `None` | 0 |

After the change the import runs and backfills normally.

## Tests

Added `expect_contracts()` to the test helpers for mocking a multi-contract `zaehlpunkte` payload, plus two cases:

- `test_get_zaehlpunkt_prefers_active_contract` — fails on `main` (returns the expired `geschaeftspartner`), passes with the fix.
- `test_get_zaehlpunkt_falls_back_to_last_match_when_none_active` — pins the existing behaviour when no contract is active.

Full suite passes (38), `flake8` clean with the CI config.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


DarwinsBuddy (approved):
lgtm

PR #367, opened 2026-09-16
fix(importer): fall back to daily values when quarter-hour data is empty

kbrozek:
## Problem

If the quarter-hour opt-in was activated only recently, the portal already reports
`granularity: QUARTER_HOUR`, but the bewegungsdaten endpoint returns nothing for role V002:

{'descriptor': {'einheit': None, 'granularitaet': 'QH', 'rolle': 'V002', ...}, 'values': []}

The importer then hits `if bewegungsdaten['unitOfMeasurement'] is None` and aborts with
"Unit of measurement is None! Aborting import...". The daily values under role V001, which
exist for years and are visible in the portal, are never requested.

Result: statistics stay empty forever, only the meter reading sensor works. Silent failure,
nothing in the log except that one warning.

## Fix

If the response is empty, retry once with `ValueType.DAY` before giving up. Once V002
starts delivering data, the fallback no longer triggers and quarter-hour import resumes
on its own.

## Testing

Tested on a Zaehlpunkt with a fresh quarter-hour opt-in: more than one year of daily values
imported successfully, energy dashboard and statistics graphs populated.


DarwinsBuddy (approved):
lgtm

DarwinsBuddy:
had to change some bits to be more robust, due to inconsistencies of the API

https://github.com/DarwinsBuddy/WienerNetzeSmartmeter/commit/4e9d00e617200bbc28d29c770733589af793c5c9

